PR #49, opened 2015-02-26
FIX Review over the rewrite rules, some mis-filed.

camfindlay:


camfindlay:
@wilr happy with these fixes?  Seems a few of the rewrites went wild. More fixes to come.

PR #55, opened 2015-06-23
Update toolbar to use SSL css from ss.org

camfindlay:
@unclecheese think this will do the trick of using SSL CSS file for the tool bar on the docs site?

### DIFF
--- a/app/_config/toolbar.yml
+++ b/app/_config/toolbar.yml
@@ -3,7 +3,7 @@ After:
   - toolbar/*
 ---
 GlobalNav:
-  hostname: 'http://www.silverstripe.org/'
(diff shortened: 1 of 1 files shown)


unclecheese:
No, unfortunately. That may be a small part of it, though. That setting is used to compare the current host name to the "primary" host name, e.g. the one that hosts the static .html file for the toolbar.

Bare minimum, I see two updates we need to make:

* `GlobalNavLink()` returns an `AbsoluteLink()`, and rather than relying on the `base_tag`, it uses a leading //, per Sam's suggestion

* Updating the external URLs in the CMS to point to // instead of `http://`.

PR #56, opened 2015-07-01
Use protocolless URLs where possible.

edlinklater:
Fixes silverstripe/silverstripe-globaltoolbar#26


ss23:
looks good, will merge when silverstripe-globaltoolbar#28 is merged

PR #57, opened 2015-07-07
Use new keyed global toolbar

edlinklater:
@unclecheese


camfindlay:
I'll deploy this out to our beta.docs site to test it.

PR #58, opened 2015-07-08
Block unstable versions from being indexed by search engines

edlinklater:
Google by default prefers to show the newest content, meaning that users are seeing 4.0 docs by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 /src/
 .DS_Store
-.idea/
 /silverstripe-cache
 /assets
 /vendor/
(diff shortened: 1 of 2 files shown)


ss23:
I think preventing it being crawled is a bad idea. Is there an alternative way to make Google prefer 3.X? See MySQL for example - they default to 5.1 in Google results at the moment.

camfindlay:
I'm ok with the very much non-stables being excluded for now. 3.2 being indexed would be good given we are in beta. If you can find another method to tell google to boost 3.1 while still indexing the others we'd take a follow up PR @ss23 ;)

edlinklater:
The ideal would be for us to generate a sitemap.xml with weights - we can use that to tell Google which to prefer. But that probably won't be simple to do here, so I think blocking the (very) unstable versions is better than nothing in the meantime.

PR #60, opened 2015-07-09
Updated changelog rewrite rule.

SpiritLevel:


SpiritLevel:
Another question!

It seems that http://docs.silverstripe.org/datefield should point to http://api.silverstripe.org/3.1/class-DateField.html

However, the api target is version dependent, which doesn't seem desirable. And, http://api.silverstripe.org/class-DateField.html doesn't default to http://api.silverstripe.org/3.1/class-DateField.html as the docs urls do. For example, 

http://docs.silverstripe.org/en/getting_started/ points to http://docs.silverstripe.org/en/3.1/getting_started/

From my naive viewpoint, http://api.silverstripe.org/ should also default to something like http://api.silverstripe.org/3.1 as the docs do. How do you want to handle the many rewrites that point to the api docs ?

:)

camfindlay:
DateField information is at http://docs.silverstripe.org/en/developer_guides/forms/field_types/datefield/

SpiritLevel:
@camfindlay That rewrite rule checker is handy :) Let's try these first. I'll do some more then squash; I wasn't working on a local version (just direct editing on GitHub). I'm now setup with a local fork so can squash.

Issue #50, opened 2015-03-02
Top 50 404 errors


camfindlay:
These are the current top 50 404 crawl errors from Google Webmaster Tools, will need to map these to appropriate places. The update in #49 should correct some of these, the rest may need to be added to the htaccess. I'll update this list as I see crawl fixes after deploying the new htaccess file.

Some are for really old versions of the docs which we do not serve up anymore (2.3), we'll need a place for these to redirect to (either to 2.4 docs, or perhaps the docs folder in github?).

The following is in order of most requested 404s:

404 http://docs.silverstripe.org/changelog
404 http://docs.silverstripe.org/installation-into-subversion
404 http://docs.silverstripe.org/modules:rssaggregator
404 http://docs.silverstripe.org/memberauthenticator
404 http://docs.silverstripe.org/modules:mssql
404 http://docs.silverstripe.org/authenticator
404 http://docs.silverstripe.org/recipes:combining_files
404 http://docs.silverstripe.org/datefield
404 http://docs.silverstripe.org/member
404 http://docs.silverstripe.org/en/topics/datamodel/
404 http://docs.silverstripe.org/treedropdownfield
404 http://docs.silverstripe.org/en/2.4/tutorials/1-building-a-basic-site/
404 http://docs.silverstripe.org/en/topics/security/
404 http://docs.silverstripe.org/en/topics/environment-management/
404 http://docs.silverstripe.org/en/topics/testing/
404 http://docs.silverstripe.org/en/2.4/tutorials/2-extending-a-basic-site/
404 http://docs.silverstripe.org/en/topics/configuration/
404 http://docs.silverstripe.org/download
404 http://docs.silverstripe.org/en/reference/permission/
404 http://docs.silverstripe.org/en/topics/controller/
404 http://docs.silverstripe.org/typography
404 http://docs.silverstripe.org/en/howto/
404 http://docs.silverstripe.org/en/topics/rich-text-editing/
404 http://docs.silverstripe.org/en/reference/modeladmin/
404 http://docs.silverstripe.org/en/3.0/tutorials/5-dataobject-relationship-management/
404 http://docs.silverstripe.org/en/2.4/tutorials/5-dataobject-relationship-management/
404 http://docs.silverstripe.org/en/2.4/tutorials/3-forms/
404 http://docs.silverstripe.org/en/3.0/tutorials/2-extending-a-basic-site/
404 http://docs.silverstripe.org/en/2.4/topics/reference/requirements/
[...]

PR #63, opened 2015-07-13
Issue 50 ( add rewrite rules for top 50 404 errors)

SpiritLevel:


camfindlay:
Not too sure about commenting out all the past rewrites, many are there for a reason based on the urls of the old docs structure.

SpiritLevel:
Understood :) But, can't we just merge these, re-run the top 50, confirm that I am getting the regexes correct for the docs.silverstripe.org server. Then, re-generate the top 50, and keep repeating this until all 404s are eliminated ? If you are worried about leaving too many 404s "live", I'll agree to attend to the list right away if you agree to keep merging and generating the new sets of top 50 ?

camfindlay:
The top 50 are over and above the previous rewrites which we must continue to account for. So any pull request would have to:

1. Retain historic rewrites (otherwise if we remove these the net total of 404s will actually rise not fall).
2. Fix the listed top 50 in the #50 after which time we would close that issue and open a new one with a new set of rewrites to tackle.

If you want to re-structure how the historic rewrites are written for the sake of tidiness that is fine as long as they still do the same job.

Ends is what matter most, not means - in this case value to community is not getting 404s when they visit the docs site from old links.

SpiritLevel:
OK. I'll put the historic rewrites back in; this will cost us in terms of many additional RewriteConds though :)

camfindlay:
Can you give me an example?

SpiritLevel:
Just doing them now. So far, the RewriteConds I already have are preventing loops! So, maybe no new ones are needed after all :)

camfindlay:
I mean an example of a loop you are dealing with and your reasoning :)

SpiritLevel:
All the loops are commented in the
      rules I added. Roughly 15 of them.
      
      On 14/07/15 11:28, Cam Findlay wrote:
    
    
      I mean an example of a loop you are dealing with and your
        reasoning :)
      —
        Reply to this email directly or view
          it on GitHub.

camfindlay:
Ok @SpiritLevel 

So with regards to this cond and rule as an example.

```
RewriteCond %{REQUEST_URI} !^(.*)/field_types/datefield
RewriteRule ^(.*)datefield/?$ /$1/en/developer_guides/forms/field_types/datefield [R=301,L]
```

The 404 we are fixing here is http://docsdomainhere/datefield only

So the rule will simply be:

```
RewriteRule ^datefield/?$ /en/developer_guides/forms/field_types/datefield [R=301,L]
```

It should only match http://docsdomainhere/datefield or http://docsdomainhere/datefield/
and the destination should not need to include the $1 since the incoming link is not providing us with any hint to a locale. In this case we default to our known /en/.

This means that (.*)/field_types/datefield won't match and cause a loop so no need for the extra RewriteCond.

You can test this over at http://htaccess.madewithlove.be/ :)

SpiritLevel:
OK. Will check RewriteConds to see which can be eliminated :)

PR #64, opened 2015-07-14
FIXED : Issue 50

SpiritLevel:
Right, let's test this!


camfindlay:
try running 

```
git rebase -i origin/master on your local branch then leave the top item as 'pick' and the second item change to 'squash'. This will rewrite the version history and squash your commits into 1. Then just force push the change to your forked branch. Some docs at http://docs.silverstripe.org/en/3.1/contributing/code/#squash-all-commits-related-to-a-single-issue-into-a-single-commit

File: .htaccess [hunk 1]
```diff
@@ -156,14 +156,67 @@
...
-
+	# RewriteRule ^framework/en/contributing/?(.*) http://doc.silverstripe.org/framework/en/trunk/contributing/$1 [R=301,L]
+	# RewriteRule ^framework/en/[\d.]*/contributing/?(.*) http://doc.silverstripe.org/framework/en/trunk/contributing/$1 [R=301,L]
+
+
+	#### Issue 50 : Top 50 404 Errors ####
+	RewriteRule ^changelog/?$ /en/changelogs [R=301,L]
+	RewriteRule ^installation-into-subversion/?$ /en/getting_started/composer [R=301,L]
+	# check target with Cam
+	RewriteRule ^modules:rssaggregator/?$ http://addons.silverstripe.org/add-ons/silverstripe/versionfeed [R=301,L]
```
camfindlay: We might be better to just drop the incoming link here on the docs home page. As this refers to an old module that now has a circular reference. See https://github.com/sminnee/silverstripe-rssaggregator which mentions that the docs for this are at the old incoming link. I don't think these docs exist on the docs site anymore.

camfindlay: Make sure when you resolve these you dont make another pull request. Just update this one.

.htaccess
```diff
@@ -156,14 +156,67 @@
...
+	RewriteRule ^memberauthenticator/?$ /en/developer_guides/security/authentication [R=301,L]
+	# check target with Cam
+	RewriteRule ^modules:mssql/?$ http://addons.silverstripe.org/add-ons/silverstripe/mssql [R=301,L]
```
camfindlay: Seems like a good destination URL.

.htaccess
```diff
@@ -156,14 +156,67 @@
...
+	RewriteRule ^en/topics/controller/?$ /en/developer_guides/controllers [R=301,L]
+	RewriteRule ^typography/?$ /en/developer_guides/customising_the_admin_interface/typography [R=301,L]
+	RewriteRule ^en/howto/?$ /en/tutorials [R=301,L]
+	RewriteRule ^en/topics/rich-text-editing/?$ /en/developer_guides/forms/field_types/htmleditorfield [R=301,L]
+	RewriteRule ^en/reference/modeladmin/?$ /en/developer_guides/customising_the_admin_interface/modeladmin [R=301,L]
+	RewriteRule ^(.*)/tutorials/5-dataobject-relationship-management/?$ /$1/tutorials/dataobject-relationship-management [R=301,L]
+	RewriteRule ^(.*)/tutorials/3-forms/?$ /$1/tutorials/forms [R=301,L]
+	RewriteRule ^en/2.4/topics/reference/requirements/?$ http://docs.silverstripe.org/en/2.4/reference/requirements [R=301,L]
+	# check target with Cam
+	RewriteRule ^modules:auth_ext_drivers/?$ https://github.com/mattclegg/silverstripe-doc-restructuring/blob/master/output/modules/auth_ext_drivers.md [R=301,L]
```
camfindlay: Maybe link to this on addons if it's there?

.htaccess
```diff
@@ -156,14 +156,67 @@
...
+	RewriteRule ^(.*)reference/cms-architecture/?$ /$1/cms_architecture [R=301,L]
+	RewriteRule ^(.*)fields/common_subclasses/?$ /$1/field_types/common_subclasses [R=301,L]
+	RewriteRule ^tabset/?$ /en/developer_guides/forms/tabbed_forms [R=301,L]
+	RewriteRule ^en/trunk/developer_guides/extending/extensions/?$ /en/developer_guides/extending/extensions [R=301,L]
+	RewriteRule ^en/topics/css/?$ /en/developer_guides/templates/requirements [R=301,L]
+	RewriteRule ^en/developer_guides/customising_the_cms/modeladmin/?$ /en/developer_guides/customising_the_admin_interface/modeladmin [R=301,L]
+	RewriteRule ^(.*)sqlquery/?$ /$1/sql_query [R=301,L]
+	RewriteRule ^en/installation/upgrading/?$ /en/upgrading [R=301,L]
+	RewriteRule ^en/developer_guides/data_model_and_orm/?$ /en/developer_guides/model/data_model_and_orm [R=301,L]
+	###  end of Issue 50  ###
```
camfindlay: No need to retain comments referring to GH issues.

.htaccess
```diff
@@ -156,14 +156,67 @@
...
 	# Anything in sapphire/en/<version>/contributing is redirected to trunk, the info in there is usually not version specific,
 	# and only updated on master (e.g. contribution and coding guidelines)
-	#RewriteRule ^framework/en/contributing/?(.*) http://doc.silverstripe.org/framework/en/trunk/contributing/$1 [R=301,L]
-	#RewriteRule ^framework/en/[\d.]*/contributing/?(.*) http://doc.silverstripe.org/framework/en/trunk/contributing/$1 [R=301,L]
-
+	# RewriteRule ^framework/en/contributing/?(.*) http://doc.silverstripe.org/framework/en/trunk/contributing/$1 [R=301,L]
+	# RewriteRule ^framework/en/[\d.]*/contributing/?(.*) http://doc.silverstripe.org/framework/en/trunk/contributing/$1 [R=301,L]
+
+
+	#### Issue 50 : Top 50 404 Errors ####
```
camfindlay: Remove this comment after you resolve the rewrites in this pull request.

.htaccess
```diff
@@ -156,14 +156,67 @@
 
 	# Anything in sapphire/en/<version>/contributing is redirected to trunk, the info in there is usually not version specific,
```
camfindlay: I wonder if we can actually drop these given they are commented out already? Thoughts @chillu ?

.htaccess
SpiritLevel: Nothing shows under auth_ext_drivers.

camfindlay:
After a final fix up I'll look to deploy this to our beta test server and run a script with the 50 incoming urls we are fixing to check the responses.

.htaccess
SpiritLevel: Done.

.htaccess
SpiritLevel: Done.

.htaccess
camfindlay: Hmmm perhaps in these "modules:*" cases maybe we have a standard to link to them on addons if they exist otherwise just send the visitor to the home of the addons site?

.htaccess
```diff
@@ -156,14 +156,64 @@
...
-	#RewriteRule ^framework/en/contributing/?(.*) http://doc.silverstripe.org/framework/en/trunk/contributing/$1 [R=301,L]
-	#RewriteRule ^framework/en/[\d.]*/contributing/?(.*) http://doc.silverstripe.org/framework/en/trunk/contributing/$1 [R=301,L]
-
+	# RewriteRule ^framework/en/contributing/?(.*) http://doc.silverstripe.org/framework/en/trunk/contributing/$1 [R=301,L]
+	# RewriteRule ^framework/en/[\d.]*/contributing/?(.*) http://doc.silverstripe.org/framework/en/trunk/contributing/$1 [R=301,L]
+
+	RewriteRule ^changelog/?$ /en/changelogs [R=301,L]
+	RewriteRule ^installation-into-subversion/?$ /en/getting_started/composer [R=301,L]
+	# See https://github.com/sminnee/silverstripe-rssaggregator. Docs no longer exist.
+	# RewriteRule ^modules:rssaggregator/?$ http://addons.silverstripe.org/add-ons/silverstripe/versionfeed [R=301,L]
```
camfindlay: Send this to home page of addons site. No need for the above comment. Once the 404 is fixed we are unlikely to revisit.

.htaccess
```diff
@@ -156,14 +156,64 @@
...
+	RewriteRule ^memberauthenticator/?$ /en/developer_guides/security/authentication [R=301,L]
+	# check target with Cam
```
camfindlay: Remove comment

.htaccess
```diff
@@ -156,14 +156,64 @@
...
+	RewriteRule ^en/reference/permission/?$ /en/developer_guides/security/permissions [R=301,L]
+	RewriteRule ^en/topics/controller/?$ /en/developer_guides/controllers [R=301,L]
+	RewriteRule ^typography/?$ /en/developer_guides/customising_the_admin_interface/typography [R=301,L]
+	RewriteRule ^en/howto/?$ /en/tutorials [R=301,L]
+	RewriteRule ^en/topics/rich-text-editing/?$ /en/developer_guides/forms/field_types/htmleditorfield [R=301,L]
+	RewriteRule ^en/reference/modeladmin/?$ /en/developer_guides/customising_the_admin_interface/modeladmin [R=301,L]
+	RewriteRule ^(.*)/tutorials/5-dataobject-relationship-management/?$ /$1/tutorials/dataobject-relationship-management [R=301,L]
+	RewriteRule ^(.*)/tutorials/3-forms/?$ /$1/tutorials/forms [R=301,L]
+	RewriteRule ^en/2.4/topics/reference/requirements/?$ http://docs.silverstripe.org/en/2.4/reference/requirements [R=301,L]
+	# check target with Cam
```
camfindlay: Remove comment

.htaccess
```diff
@@ -156,14 +156,61 @@
...
+	RewriteRule ^memberauthenticator/?$ /en/developer_guides/security/authentication [R=301,L]
+	RewriteRule ^modules:mssql/?$ http://addons.silverstripe.org/add-ons/silverstripe/mssql [R=301,L]
+	RewriteRule ^authenticator/?$ /en/developer_guides/security/authentication [R=301,L]
+	RewriteRule ^recipes:combining_files/?$ /en/developer_guides/templates/requirements [R=301,L]
+	RewriteRule ^datefield/?$ /en/developer_guides/forms/field_types/datefield [R=301,L]
+	RewriteRule ^member/?$ /en/developer_guides/security/member [R=301,L]
+	RewriteRule ^en/topics/datamodel/?$ /en/developer_guides/model/data_model_and_orm [R=301,L]
+	RewriteRule ^treedropdownfield/?$ /en/developer_guides/forms/field_types/common_subclasses [R=301,L]
+	RewriteRule ^(.*)/tutorials/1-building-a-basic-site/?$ /$1/tutorials/building-a-basic-site [R=301,L]
+	RewriteRule ^en/topics/security/?$ /en/developer_guides/security [R=301,L]
```
camfindlay: Might pay to check some of these against the historic rewrites. There is already this rule there just not taking into account the / on the incoming url. See https://github.com/silverstripe/doc.silverstripe.org/blob/master/.htaccess#L133

.htaccess
SpiritLevel: What about line 194 that refers request for auth_ext_drivers to github ?

.htaccess
SpiritLevel: I tested these and some required the /? while some didn't.

The one above on the old tutorials is like this.

camfindlay: yip the /?$ simply means the last character of the match can be / or no /

camfindlay: Cross ref your work with the historic rewrites, fix up the historics if they just require a minor improve to get them to correct the top 50 item they relate to. Only add new rules for new incoming links that result in 404s.

SpiritLevel: Other historical rules don't include /? but seem to work. I suppose this is because the original request luckily did not include the trailing slash ?

camfindlay: yeap, best to cater for either as we can't control whether the incoming link includes the slash or not.

camfindlay: It might be a lot of the top 50 get cleared with this simple improve to many existing rules.

Getting there... ;) Looking forward to getting a final PR and merge on this (hope you have learned a bunch!).

PR #67, opened 2015-07-17
FIXED : Issue 66

SpiritLevel:
removed extra comment mark


camfindlay:
I'll accept this in and might pay to review over to make sure there aren't any further loops where the destination url matches the rewrite.

SpiritLevel:
OK. Will do :)

PR #69, opened 2015-07-19
Issue 66 : commented out three extraneous RewriteConds and tightened …

SpiritLevel:
…a RewriteRule to avoid redirect loops

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -28,9 +28,9 @@
 
 	# Redirect legacy URLs (assumes we're not redirecting other assets, for performance reasons)
 	# Note: Just works on top-level domains, not if the webroot is in a subfolder
-	RewriteCond %{REQUEST_URI} ^(.*)$
-	RewriteCond %{REQUEST_FILENAME} !-f
-	RewriteCond %{REQUEST_URI} !\.(css|gif|ico|jpg|js|png|swf|txt)$
+	# RewriteCond %{REQUEST_URI} ^(.*)$
+	# RewriteCond %{REQUEST_FILENAME} !-f
+	# RewriteCond %{REQUEST_URI} !\.(css|gif|ico|jpg|js|png|swf|txt)$
 	#RewriteRule ^execution-pipeline$ http://doc.silverstripe.org/framework/en/reference/ [R=301,L]
 
 	RewriteCond %{REQUEST_URI} !(3.0|2.4)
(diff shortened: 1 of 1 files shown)


camfindlay:
Seems you have included 2 commits here. Is the readme one for another pull request?

SpiritLevel:
Oopsie...that second commit should have been in another branch. Sorted with a revert and push :)

camfindlay:
@SpiritLevel I'll accept this and deploy to our beta test site. Can you then test some urls and ensure nothing is broken. Especially look for infinite redirection loops. 

Something else we need to look at is that before version 3.1 (3.0/2.4) the structure of the docs where different. I've find case where we push /3.0/ and /2.4/ incoming links into the new structure of the docs (which results in a 404). Let's address this as a separate issue.

.htaccess
```diff
@@ -28,9 +28,9 @@
 
 	# Redirect legacy URLs (assumes we're not redirecting other assets, for performance reasons)
 	# Note: Just works on top-level domains, not if the webroot is in a subfolder
-	RewriteCond %{REQUEST_URI} ^(.*)$
-	RewriteCond %{REQUEST_FILENAME} !-f
-	RewriteCond %{REQUEST_URI} !\.(css|gif|ico|jpg|js|png|swf|txt)$
+	# RewriteCond %{REQUEST_URI} ^(.*)$
```
camfindlay: Reason for these to be commented out?

SpiritLevel: Well, Since Line 34 was originally commented out, Lines 31-33 do nothing except burn up resources checking the conditions. I am a big fan of removing stuff that isn't needed ;)

camfindlay: ok all good.

SpiritLevel:
@camfindlay OK. Let me know when it is on the dev server and I'll test all of them. If you create the new issue, I'll jump onto it and try to sort it. Would it be beneficial to generate another Top 50 ?

camfindlay:
on the server at beta.docs.silverstripe.org (hopefully you can reach that). I have started a new top 50 though I think another issue to sort first is the 3.0 and 2.4 docs were in a different structure that the 3.1 docs. This means any redirect for example en/2.4/developer_guides/... will 404 because there was no "developer_guides" in the old docs structure. You can see the older structure if you navigate to these areas of the docs.

SpiritLevel:
OK. Tested all redirect on beta.docs.ss.org/ and none loop or end up on the "Page not found" page :)

Let me know when the next 50 are ready to tackle. I'll look at the urls for 3.0 and 2.4.

PR #72, opened 2015-07-20
Force to SSL only on Live server

camfindlay:


camfindlay:
Cheers will look at deploy today

Issue #80, opened 2015-10-15
Cosmetic ideas to improve UX


SpiritLevel:
(1) If the contrast between the white version tab (on left) and the gray of the third band from top containing archive drop down were greater, it would be visually easier to see the version. Perhaps a title like "Version:" on the left of the version tabs too ?

(2) To be consistent, the title of third band could perhaps be "SilverStripe Developer Documentation" or more like the "SilverStripe 3.2 API Docs" and display version in title like this: "SilverStripe 3.1 Developer Documentation" ? Same with Userhelp ?

(3) Perhaps the contents of the third band when viewing an archived version could be adjusted to make it more explicit that you are viewing an archived version ? 

(4) Most people will be after the stable version of the docs. I was wondering if it might be a good idea to make the stable version prominent, collect the future/dev versions in one group called "Dev", and put the non-archived but older versions in one group called "Legacy", or something along those lines...

SpiritLevel:
The more I look at the API docs, the more I like the layout. It would be nice to blend the style of the API docs with the developer and userhelp docs and make all three styled consistently.

SpiritLevel:
Another thought: replace "Archive" with "Archived Versions", possibly with some comment that these are no longer being supported ?

PR #83, opened 2015-10-19
Cosmetics (see #80)

SpiritLevel:


camfindlay:
thanks @SpiritLevel will look at these in upcoming hackday. FYI, we have a public hack coming up you might want to attend, check the Wellington meetup group.

camfindlay:
Can you give me a quick screenshot as to what this looks like locally?

SpiritLevel:
screenshot from 2015-10-22 09 47 53

Font size is a bit big though.

camfindlay:
Perhaps the version label should be more subtle? Also how does this look on mobile?
We had the versions lining up with the side nav

SpiritLevel:
Hi Cam....just looking at docs.silverstripe.org on mobile now and there is no side nav...also, the top menu and version numbers are microscopic in portrait! Very hard to read. Landscape is tolerable. Same with items under "SilverStripe Concepts" and "Contributing to SilverStripe". The archive dropdown at  top and buttons at the bottom are microscopic too. If you want docs.silverstripe.org to have a responsive design then it appears a bit of layout work is needed!

SpiritLevel:
And, it would be nice to get https://github.com/silverstripe/silverstripe-docsviewer/pull/74 in :)

camfindlay:
Yep the top nav needs responsive work across all the docs related sites (it's pulled in from a statically generated copy from the main silverstripe.org website.

Can look to add these labels if you think it makes it clearer the numbers are version numbers (I excluded adding a "Versions" as I though it was a bit cleaner.

SpiritLevel:
Personally, I like trivial things to be crystal clear so that I don't have to spend any time at all thinking about what they mean ;) So I'd put in a heading like "Versions" or "Version" or "Documentation Versions" or something in the title "SilverStripe Developer Documentation Version 3.2"...anything to make it clear that one can select which version of the docs to read.. Also, once a version is selected, one should be able to see the selected version from wherever they are in the docs and the HOME button should keep one within that version choice, which is the reason for https://github.com/silverstripe/silverstripe-docsviewer/pull/74. 

@camfindlay What is happening with docsviewer anyway ? There are few old PRs and lots of red X's :(

camfindlay:
Docviewer unit tests are tracking against the 3 branch of core (future 3.3) hence the red Xs. We need to perhaps update the travis CI config.

SpiritLevel:
Would the config file be .travis.yml ? And are the updates needed those given in 
https://github.com/silverstripe-labs/silverstripe-travis-support#travis-setup ?

PR #84, opened 2015-10-19
Cosmetics (see #80 and #83)

SpiritLevel:


camfindlay:
Ideally you would have had this css alongside the previous PR, but I'll let you away with it this time...

SpiritLevel:
Too kind! I know...I just quickly chucked both in using the "Edit This Page" button. I'll get my docs repo sorted...

PR #91, opened 2015-11-20
fix rewrite rules for old versions of documentation

dvp10:
fix rewrite rules to stop redirecting to new url scheme on 2.4 and 3.0 versions of documentation

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -38,140 +38,243 @@
 	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/topics/testing/create-functional-test$ /$1/topics/testing/creating-a-functional-test [R=301,L]
 
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/?$ /$1/developer_guides [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/cache-control/?$ /$1/developer_guides/performance/caching [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/cms-alternating-button/?$ /$1/developer_guides/customising_the_admin_interface/how_tos/cms_alternating_button [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/cms-formfield-help-text/?$ /$1/developer_guides/customising_the_admin_interface/how_tos/cms_formfield_help_text [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/csv-import/?$ /$1/developer_guides/integration/csv_import [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/customize-cms-menu/?$ /$1/developer_guides/customising_the_admin_interface/how_tos/customise_cms_menu [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/customize-cms-pages-list/?$ /$1/developer_guides/customising_the_admin_interface/how_tos/customise_cms_pages_list [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/customize-cms-tree/?$ /$1/developer_guides/customising_the_admin_interface/how_tos/customise_cms_tree [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/dynamic-default-fields/?$ /$1/developer_guides/model/how_tos/dynamic_default_fields [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/extend-cms-interface/?$ /$1/developer_guides/customising_the_admin_interface/how_tos/extend_cms_interface [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/gridfield-rowaction/?$ /$1/developer_guides/forms/how_tos/create_a_gridfield_actionprovider [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/grouping-dataobjectsets/?$ /$1/developer_guides/model/how_tos/grouping_dataobject_sets [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/navigation-menu/?$ /$1/developer_guides/templates/how_tos/navigation_menus [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/pagination/?$ /$1/developer_guides/templates/how_tos/pagination [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/howto/simple-contact-form/?$ /$1/developer_guides/forms/how_tos/simple_contact_form [R=301,L]
 
 	RewriteRule ^en/installation/?$ /en/getting_started/installation [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/installation/common-problems/?$ /$1/getting_started/installation/common_problems [R=301,L]
 	#
 	RewriteCond %{REQUEST_URI} !en/getting_started/installation/composer
 	RewriteRule ^(.*)/installation/composer/?$ /$1/getting_started/composer [R=301,L]
 	RewriteRule ^en/getting_started/installation/composer/?$ /en/getting_started/composer [R=301,L]
 	#
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/installation/from-source/?$ /$1/getting_started/installation [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/installation/lighttpd/?$ /$1/getting_started/installation/how_to/configure_lighttpd [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/installation/mac-osx/?$ /$1/getting_started/installation/mac_osx [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/installation/nginx/?$ /$1/getting_started/installation/how_to/configure_nginx [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
 	RewriteRule ^(.*)/installation/nginx-hhvm/?$ /$1/getting_started/installation/how_to/setup_nginx_and_hhvm [R=301,L]
+	RewriteCond %{REQUEST_URI} !(3.0|2.4)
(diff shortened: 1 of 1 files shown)


SpiritLevel:
Yikes! That .htaccess file is getting mighty big; there is almost a rewrite rule for every individual link. This will make the 2.4 and 3.0 docs accessible, so would be a good quick fix, but it seems like there should be a more concise way to set all this up :)

dvp10:
Any RewriteCond work only for first next RewriteRule - i am only add Conditions for all rules starting with wildcard "^(.*)/"...
Another solution may be change wildcard "^(.*)/" to "^(.*/3.[1-9])/" for all matching rules without RewriteCond

SpiritLevel:
Hi @dvp10.

Yes, it would be really handy if a single RewriteCond statement could be applied to a block of RewriteRule statements.

I think the older docs have been placed within the newer web page framework of doc.silverstripe.org and the menu items on the left, among other things, need to get sent to the right places. Also, there was a name change from "sapphire" to "framework". I'd like to clean up these rules. Maybe @camfindlay has some suggestions on how to proceed ?

dvp10:
If I understand correctly, the purpose of the RewriteRule to redirect from a very old url like "/sapphire/en/topics/configuration" to corresponding new location for current stable version of Silverstripe.
This old url has no version number and if you change the prefix from "^(.*)/" to "^([^0-9]*)/" in RewriteRules, you don't need any RewriteCond to exclude urls for specific version of Silverstripe.

dhensby:
I think you can do [OR] to save declaring the condition every time

camfindlay:
Just some back story incase this changes the approach to these rewrites...

Versions 2.4 and 3.0 were in an older structure

Between version 2.4 and 3.0 the name of the framework changed from sapphire to framework (it is this repository that holds the markdown documentation for developers).

Version 3.1 WAS also in this older structure until there was some community work on improving this, this was officially launched at the beginning of this year. This is where the most complex rewrites are as we have to map the old structure to new. This is the reason the htaccess is the size it is (for the conversion that occurred during 3.1 I don't see much other way around this, a lot got moved). This new structure was not ported back to the older versions as they were either end of life or on their way towards EOL. These older versions should simply continue to display as they did (I would hope these version don't require much in the way of rewrites).

Version 3.2 is out now, this is the current stable. This shouldn't need any rewrites (the applies to dev versions of 3.3. and 4.x).

If you supply a link to the docs WITHOUT a version number it will redirect to the current stable version docs (which allows us to link to the docs from other places without having to change the links every time a version is released).

Hope that clears up some of the "why" things are and might help move forward with some rewrite improvements (which may have been difficult without the back story).

camfindlay:
Something else we could look at here is using a technique called skipping... 

So in the case of skipping a bunch of rules if it includes 2.4 or 3.0 in the REQUEST_URI
Example if it includes 3.0 or 2.4 then skip 10 (or X) rewriteRules ahead (the ones that should only apply if it's 3.1): 

RewriteCond %{REQUEST_URI} (3.0|2.4)
RewriteRule  .?  -  [S=10]
Rules
Rules
Rules
...

Though this would also apply then to 3.2+ which we also might not want. Perahsp we should be specific and say, "these rules apply to 3.1 only" ?

SpiritLevel:
Yes....I'd like to organise the rules exactly as you suggest in your last statement, based on version number somehow like this:

if version = 2.4 then {
   .. apply only 2.4 rewrites
} else if version = 3.0 {
   .. apply only 3.0 rewrites
} else if version = 3.1 {
  .. apply only 3.1 rewrites
}
...
} else if no version in url {
 .. apply version neutral rewrites
}

I suppose all the trouble will be with the version neutral rewrites :( At least this would partially tame that huge mess of rules! I'll have a look at [OR] and skipping as well as the other rewrite rules to see what might be the best way to implement. 

@camfindlay Do you want to merge @dvp10's PR just to get the older docs functioning quickly and I'll follow up with the above proposal ?

camfindlay:
Is there a good way to prove that these rules work as intended without impacting other rules?

Perhaps you can run a tool like http://peacockmedia.software/mac/integrity/free.html over your local copy of the site, then implement the rewrites here, run again and look for either improved or worse 404s. Rewrites have been an issue for some time, we've been throwing patch after patch at it. Without too much testing of the effects.

SpiritLevel:
OK, will dive in.

SpiritLevel:
Look what I found! As long as SS is using apache 2.4+, we can do something like this:

<IfModule mod_rewrite.c>
	RewriteEngine On
	<If "%{REQUEST_URI} =~ /test1.url/">
	    RewriteRule (.*) http://spiritlevel.nz [R=301,L]
	</If>	
	<Elseif "%{REQUEST_URI} =~ /test2.url/">
	    RewriteRule (.*) http://google.com [R=301,L]
	</Elseif>
	<Elseif "%{REQUEST_URI} =~ /test3.url/">
	    RewriteRule (.*) http://silverstripe.com [R=301,L]
	</Elseif>
	<Else>
	    RewriteRule (.*) http://silverstripe.org [R=301,L]
	</Else>
</IfModule>

camfindlay:
@dvp10 @SpiritLevel @dhensby - just checking in with this PR, would you guys like to have this merged or look at addressing the rewrites more holistically?

SpiritLevel:
Hi @dvp10 @SpiritLevel @dhensby  

<b>Here is what I have done so far:</b>

(1) on my local copy of doc.silverstripe.org's .htaccess, I turned off all rewrite rules except the one that sends urls to main.php.

(2) I added the .htaccess workaround that deals with mod_dir and mod_rewrite collisions (seehttps://github.com/silverstripe/silverstripe-installer/blob/3.2/.htaccess#L35 ). Note that this workaround is not needed for newer versions of apache (I believe 2.4.10+;  I have the reference). And, for earlier versions, there is a better workaround that retains the directory index behaviour..if we want to keep the directory indexing we just need to know if the docs' site is using and earlier version of apache.

(3) I ran several link checking tools and restricted the checks to the directory doc.silverstripe.org for each version branch post-2.4,3.0,3.1,3.2,3.3. This resulted in lists for each version consisting of purely internally generated 404s.

(4) I have started to correct these internal 404s. See:
https://github.com/silverstripe/silverstripe-framework/pull/4807
https://github.com/silverstripe/silverstripe-framework/pull/4808

There aren't all that many errors. There are roughly 30 errors in 2.4, 50 errors in 3.0, 60 in 3.1, 70 in 3.2 and 3.3. I have some questions about the best way to submit PRs for fixing these errors since they change from version to version. Generally, they are cumulative. Occasionally there is a fix upon a fix upon a fix....all complicated by the multiple patches of rewrite rules. 

<b>The approach I propose and have started to take: </b>

(i) Comment out all rewrite rules (except for the main.php handler rule and the mod_dir workaround).
(ii) Correct links from within docs.silverstripe.org to within docs.silverstripe.org and merge them.
(iii) Correct links from within docs.silverstripe.org to outside and merge them.

Steps (ii) and (iii) deal with 404s due to requests originating from urls <b>WITHIN</b> the documentation. After this work, there should be no need for <b>ANY</b> rewrite rules due to internal urls. At this stage, I propose we merge and deploy the .htaccess file with no extra rewrite rules.

(iv) merge and deploy .htaccess with NO rewrite rules (except for the main.php handler rule and the mod_dir workaround, if needed).
(v) monitor 404s that result from requests originating outside of the docs. As a courtesy to those that have their own links/bookmarks to the docs, we add a minimal set of rewrite rules as needed.

As far as this PR goes, I would say that it can be merged right away since it just blocks rules with incorrect behaviour. After you guys give some feedback about my proposed approach, and after I sort out all the internally generated 404s, we can replace the .htaccess with the minimal one in step (iv). 

What do you all think ?

SpiritLevel:
@camfindlay it looks like [S], skip, can be used to mimic if-elseif-else if we find if-elseif-else don't work as desired, but this is an awkward construct.

@dhensby I checked up on [OR] and it looks like it only takes a sequence of RewriteCond directives, which involves an implicit "AND" and changes them to explicit "OR"s...

:)

dhensby:
I think you're approach sounds reasonable although removing all the re-writes for a time until we start to see external 404s is a bit painful for users.

I think lots of those rewrites are there because of external links.

@camfindlay do you have webmaster tools for the docs site? It'll tell us most of the external 404s that we get...

dhensby:
In terms of accumulative fixes, I think you'll just have to duplicate the work across branches - that's the easiest way. Otherwise you'll be waiting for merge-ups and stalling for a long time.

SpiritLevel:
Alrighty. 

So, between (iii) and (iv), we generate a list of external 404s and set up rules for them. THEN, deploy the .htaccess.  

And, submit changes to each and every branch rather than waiting for merge-ups.

camfindlay:
@dhensby @SpiritLevel we did have webmaster tools, however it seems to have ceased working, I'll look into why (very annoy!).

SpiritLevel:
I see that www.silverstripe.org is running cloudfare-nginx....can someone tell me the nginx version ? Then I can check if the if-elseif-else control structure can be used. Also, if we want directory index behaviour, knowing the version will help determine whether or not we need we need this workaround:

http://serverfault.com/questions/607207/directoryindex-not-working-properly-for-rewritten-url

camfindlay:
cloud flare is a WAF and CDN service, I don't think we can edit their nginx ;) This is not used on the docs site fyi.

SpiritLevel:
Oh...I did <pre>curl --head www.silverstripe.org</pre> and the server is returned as cloudfare-nginx. 

And, both doc.silverstripe.org and docs.silverstripe.org have been 301'd....

So...where are the SS docs hosted and what is version of the server being run?

camfindlay:
I'll find out version of apache running and let you know.

SpiritLevel:
Excellent :)

camfindlay:
ok, we are on apache 2.2.16 and php 5.3.3 (not the latest, that's a separate issue).

camfindlay:
hopefully that can direct your approach @SpiritLevel

SpiritLevel:
Perfect :) I'll first do all the internal links on the 2.4 docs. Do you want the directory indexing to function or are you happy to have it disabled, just like in the installer ?

camfindlay:
Either way is fine, I can create a branch for us to test on and deploy if you like while we are poking at the htaccess

SpiritLevel:
Yes, sounds like a really good idea :) Can never test enough!

SpiritLevel:
@camfindlay There are references to modules in the 2.4 docs. When I search addons for 2.4 modules I typically can't find them and get the message that they may be on http://www.silverstripe.org/modules. For example GenericDataAdmin and RelatedDataEditor.

Anyhow, do you have these 2.4 modules somewhere else I can link to or should I just make the links to them inactive ?

SpiritLevel:
OK, all internally generated 404s sorted: https://github.com/silverstripe/silverstripe-framework/pull/4815

I started on links in the 2.4 docs that go to "external" sites and there are a number of issues:

* markdown elements referring to classes of the form [api:DataObjectSet] are parsed to call http://api.silverstripe.org/search/lookup/?q=DataObjectSet but some work and some don't. And, the occurs over different versions. How to fix ?
* markdown elements referring to class methods of the form [api:DataObjectSet->groupBy()] are not   working correctly either, also over different versions. How to fix ?
* what do you want to do about links to open.silverstripe.org ?
* what do you want to with translate.silverstripe.org ?

There are lots of the first three issues.

camfindlay:
I'm going to merge this in and test today. We can look to reorganise afterward however this fixes up a few issues when accessing 2.4 and 3.0 docs.

SpiritLevel:
Excellent. After  silverstripe/silverstripe-framework#4816 and silverstripe/silverstripe-framework/pull/4808 are merged, I'll do some more links tidying :)

camfindlay:
@SpiritLevel have emailed you... if you can test this rewrite and let me know I'll look to push to production today.

camfindlay:
This has been pushed to production.

SpiritLevel:
Good! I'll follow up with more link tidying.

SpiritLevel:
@camfindlay Just checking in :)  I'll resume cleaning up internal links in the next few days. Are you able to get the top 100 (or top 200 or top 300) 404s arising from external requests only ? As @dhensby pointed out, it would be painful for users if we completely eliminate all rewrites rules and then after add a minimal set back in. So, in order to balance the cleaning up of the rewrite rules against maintaining rewrite rules for commonly accessed links, I propose we start with a decent sized list (100, 200, 300 or so) of externally generated 404s and I'll work out a <b>minimal</b> set of rewrite regexs for these.

PR #92, opened 2015-11-20
MINOR: move favicons to themes/docs/favicons/ folder

SpiritLevel:
See #90.

PR #94, opened 2015-11-23
fixed MS browser config schema

SpiritLevel:
Oopsie...forgot to do this as part of favicon cleanup. See https://github.com/silverstripe/doc.silverstripe.org/pull/92


dhensby:
@SpiritLevel was this not needed or merged somewhere else?

SpiritLevel:
@dhensby These changed were all merged in :) This may have been me deleting forks again in my impatience...bad habit...will change my ways.

PR #95, opened 2015-11-24
changed hostname in toolbar.yml so toolbar works on local installs

SpiritLevel:


SpiritLevel:
See also https://github.com/silverstripe/userhelp.silverstripe.org/pull/17#

camfindlay:
Will be leaving this as is for now, will rethink best way to make this locally more accessible. This assumes you'll always want to use the live toolbar (which when we do work on the toolbar module causes havoc). You ok with me closing this for now @SpiritLevel ?

SpiritLevel:
@camfindlay Sure...I'll just use .gitignore and keep my toolbar.yml changes. :)

PR #104, opened 2016-02-09
Feature/platform

mandrew:
updated lucene base path, added repos yml, added crontask and buildtask, ignore lucene build folder


SpiritLevel:
Oh, I suppose README.md will need to be updated as a result of these changes :)

app/_config.php
```diff
@@ -58,9 +58,7 @@
 ));
 
 // Set shared index (avoid issues with different temp paths between CLI and web users)
-if(file_exists(BASE_PATH . '/.lucene-index')) {
-	Config::inst()->update('DocumentationSearch', 'index_location', BASE_PATH . '/.lucene-index');
-}
+Config::inst()->update('DocumentationSearch', 'index_location', BASE_PATH . '/assets/.lucene-index');
```
assertchris: Why is there here, and not in yml?

app/code/UpdateTask.php
```diff
@@ -0,0 +1,118 @@
...
+    {
+        return ASSETS_PATH;
+    }
+
+    /**
+     * @param string $message
+     */
+    private function printLine($message)
+    {
+        $this->eol = Director::is_cli() ? PHP_EOL : "<br>";
```
assertchris: Why is this being cached if it is only ever used in the next line? There isn't even a local property defined for this...

PR #105, opened 2016-02-09
stripped down .htaccess file for testing links and rewrite rules

SpiritLevel:
@camfindlay I put this in develop instead of master. Can you put this on uat for testing the api: and other links ? Thanks....

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -12,311 +12,311 @@
 <IfModule mod_rewrite.c>
 	RewriteEngine On
 
-	# Legacy rewrite from sapphire/ to framework/ namespace
-	RewriteCond %{REQUEST_FILENAME} !-f
-	RewriteCond %{REQUEST_URI} !^framework/main\.php
-	RewriteRule ^sapphire/?(.*) http://doc.silverstripe.org/framework/$1 [R=301,L]
-	
-	# Rewrite framework to the new base url
-	RewriteCond %{REQUEST_URI} ^/?framework/en/
-	RewriteRule ^framework/en/(.*)$ /en/$1 [R=301,NC,QSA,L]
-
-	# DokuWiki rewrite rules: Need to happen before other rules in order to redirect /doku.php?id=<pagename>
-	# to /pagename, which can then be matched by the legacy rewrite rules further down
-	RewriteCond %{QUERY_STRING}          ^(\bid\b=([^&]*)&?(.*)?)
-	RewriteRule ^doku.php$                http://doc.silverstripe.org/%2?%3 [R=301,L]
-
-	# Redirect legacy URLs (assumes we're not redirecting other assets, for performance reasons)
-	# Note: Just works on top-level domains, not if the webroot is in a subfolder
-	# RewriteCond %{REQUEST_URI} ^(.*)$
+	# # Legacy rewrite from sapphire/ to framework/ namespace
 	# RewriteCond %{REQUEST_FILENAME} !-f
-	# RewriteCond %{REQUEST_URI} !\.(css|gif|ico|jpg|js|png|swf|txt)$
-	#RewriteRule ^execution-pipeline$ http://doc.silverstripe.org/framework/en/reference/ [R=301,L]
-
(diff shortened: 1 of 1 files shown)


camfindlay:
Hey @SpiritLevel I'll have to look at this after the current work we're doing on migrating the docs sites to SilverStripe Platform.

SpiritLevel:
@camfindlay All good :) The only thing I did was comment out all rewrite rules except those at the bottom that send requests to be handled by main.php. Recall that I am aiming for a minimal set of rewrite rules. The first step is to fix all broken internal and external links, and this requires that all rewrites be turned off.
Once the links are fixed, then we can think about what rewrite rules to put in place for requests coming from outside SS, to help people find what they are looking for.

SpiritLevel:
@camfindlay So, I have fixed all internal and external links in all vesions <= 3.2. Will these be merged up ? 

The next step is to clear up .htaccess. How do you want to do this? I'm inclined to delete almost all the rewrites. Then, monitor 404s and put in rewrites strictly as needed. We certainly don't want to leave it the way it is. 

Related: I recall @chillu say something about being able to go from one tree node in docs of one version to the equivalent tree node in another version. This would require a mapping between the trees of all versions, which would be a pain to maintain, and is essentially what all the rewrite rules are trying to do. How important is it have the ability to do this ?

camfindlay:
I still have reservations about this as it will create a lot of 404 for active users coming from external links.

Is there a better approach here we could take? Perhaps do it in smaller blocks?

SpiritLevel:
OK, how about we cycle through the following steps:

(1) remove 5 or 10 rules from .htaccess and record which ones were removed
(2) wait a week
(3) generate a list of the top 500 404s
(4) if any 404s are due to any rules removed in any prior cycling through (1), we reintroduce any needed rule but make sure it has a regex with the minimum needed matching scope.

Why don't you generate a top 500 list right now and send me the results as a CSV file, like last time. I'll go through it. Then we can see what is currently happening, and make sure that in (1) we don't remove rules we know are being used. 

This is a slow steady approach that minimises inconvenience to those relying on the rewrite rules.

camfindlay:
+1 good strategy. I'll get you some 404 data.

SpiritLevel:
Excellent :)

dhensby:
what's the motive behind removing so many 301s?

SpiritLevel:
@dhensby I'm not sure anymore !! :P  What might be the motive for wanting to keep so many 301s ? ;)

One main goal of this effort was to reduce the volume of 404s.

Another goal was to reduce the total set of rewrite rules so that they are easier to understand and maintain. The issue is how to reduce the set.

There are "legacy" 301s that will likely have to be kept. For instance, references to "sapphire" perhaps need to be kept. I have seen some say that they would like to be able to quickly go from one page in one version of the docs to the corresponding page in another version of the docs, without having to go via the top of the document tree. The ability to do this would require one to define complete mappings between the document trees making up each version of the docs. The existing rewrite rules are an incomplete attempt to define such mappings. Personally, I think trying to set up and maintain mappings between versions of the docs would be a huge waste of time. I certainly wouldn't want to have to do it! Instead, I think that any 301 that has been set up to map between different versions of the docs should be removed so that things can be kept simple.

dhensby:
Hmmm. We should maintain 301s for legitimate old urls. I don't think we should map old to new versions here, that should probably be done as part of the application layer, but if there's no way to do that currently, I don't think we should remove them

dhensby:
@SpiritLevel you really need to stop deleting your forks because it closes all your PRs :(

SpiritLevel:
Hey @dhensby ... I intentionally closed my PRs, which I did first, then I deleted my fork. But, you're right that I shouldn't have to delete my fork...sometimes it is just easier to do that than fix up any messes that I make!

dhensby:
Oh. Ok. I'm happy to run you through how to fix up messes in git :)

SpiritLevel:
Thanks mate :)

PR #113, opened 2016-02-16
Update feedback module to 0.2.2

camfindlay:


camfindlay:
@mandrew peer and merge plz

PR #116, opened 2016-02-17
.htaccess rewrite rule, dev flag, README.md, toolbar.yml

SpiritLevel:
I think you meant to put ".org" instead of ".com" on the RewriteCond.

Also, I don't think you need the "s?" in the RewriteCond: you only want to redirect doc.silverstripe.org to docs.silverstripe.org, and just let docs.silverstripe.org pass through ?

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -30,7 +30,7 @@ ErrorDocument 500 /assets/error-500.html
 	RewriteEngine On
 
 	RewriteCond %{HTTP_HOST} ^docs?.silverstripe.com$ [NC]
-    RewriteRule ^(.*)$ https://docs.silverstripe.org/$1 [L,R=301]
+	RewriteRule ^(.*)$ https://docs.silverstripe.org/$1 [L,R=301]
 
 	# Legacy rewrite from sapphire/ to framework/ namespace
 	RewriteCond %{REQUEST_FILENAME} !-f
--- a/app/_config/toolbar.yml
+++ b/app/_config/toolbar.yml
@@ -13,7 +13,7 @@ Only:
   environment: dev
 ---
 GlobalNav:
-  hostname: '//ssorg.dev/'
+  hostname: '//www.silverstripe.org/'
   css_path: '/themes/ssv3/css/toolbar.min.css'
   snippet_path: '/assets'
   use_localhost: false
--- a/app/code/RefreshMarkdownTask.php
+++ b/app/code/RefreshMarkdownTask.php
@@ -94,7 +94,7 @@ class RefreshMarkdownTask extends BuildTask
         $this->printLine("cloning " . $remote . "/" . $branch);
 
         chdir("{$path}/src");
-        exec("git clone -q https://github.com/{$remote}.git {$folder}_{$branch} --depth 1 --branch {$branch} --single-branch");
+        exec("git clone -q https://github.com/{$remote}.git {$folder}_{$branch} --branch {$branch} --single-branch");
 
         chdir("{$path}/src/{$folder}_{$branch}");
     }
@@ -106,6 +106,10 @@ class RefreshMarkdownTask extends BuildTask
      */
     private function cleanRepository(array $repository)
     {
+	
+	// if the dev=1 flag is used when invoking RefreshMarkdownTask the git repos are kept.
+	if( isset($_REQUEST['dev']) && $_REQUEST['dev'] ) return;
+
         $paths = array_merge(glob("*"), glob(".*"));
 
         foreach ($paths as $path) {
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 # doc.silverstripe.org
 
-This is the source code powering http://docs.silverstripe.org. It
-primarily consists of the SilverStripe
+This is the source code powering http://docs.silverstripe.org and primarily consists of the SilverStripe
 [framework](https://github.com/silverstripe/silverstripe-framework)
 and the [docsviewer](https://github.com/silverstripe/silverstripe-docsviewer)
 module with minimal configuration.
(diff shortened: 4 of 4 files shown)


SpiritLevel:
I updated README.md to reflect the current state of things. I intend to look at putting a flag in `RefreshMarkdownTask` to prevent the deletion of the git repos. Also, there is a bit of redundancy in the YAML files (ie: docsviewer.yml and docs-repositories.yml) which should be eliminated if possible. Lastly, I think the two `sake` commands should be combined into a single command, and perhaps `RebuildLuceneDocsIndex` could be renamed at the same time to something like `RebuildDocsIndex`; whether or not Lucene is used is purely an internal implementation issue and should not explicitly be in the name of the task. :)  Just my opinions!

camfindlay:
1) I meant to put .com
2) RebuildLuceneDocsIndex is part of docsviewer and should be it's own task

SpiritLevel:
@camfindlay 

1) Thought so ;)

2) Yup. But, for installing docs, it would be nice to have a single `sake RefreshDocs flush=1` command that runs `RefreshMardownTask` and `RebuildDocsIndex`. Just a thought.

.htaccess
```diff
@@ -29,8 +29,8 @@ ErrorDocument 500 /assets/error-500.html
 <IfModule mod_rewrite.c>
 	RewriteEngine On
 
-	RewriteCond %{HTTP_HOST} ^docs?.silverstripe.com$ [NC]
-    RewriteRule ^(.*)$ https://docs.silverstripe.org/$1 [L,R=301]
+	RewriteCond %{HTTP_HOST} ^doc.silverstripe.org$ [NC]
```
camfindlay: Leave this as is please :D

camfindlay:
Happy to leave the scripts as two separate ones for now. More pressing matters to address.

SpiritLevel:
OK...just add dev flag and set .htaccess RewriteCond to include the "s?".

SpiritLevel:
Oh, I noticed an issue when running RefreshMarkdownTask in a fresh clone & composer install of doc.silverstripe.org. The assets/src/framework_* folders are not initially present and so DocumentationManifest complains about their absence. So, some kind of manifest refresh needs to be done but I am not sure of the best way to do this...being a newbie!

SpiritLevel:
I'll update README.md

camfindlay:
updating the manifest is just calling a flush

SpiritLevel:
OK...README.md updated as well as toolbar.yml :)

SpiritLevel:
Do you mean just including the `flush=1` flag ? I'm a bit confused since this didn't work and I can't see anywhere in `RefreshMarkdownTask` where the manifest is updated. Will try again and look a it bit more...

SpiritLevel:
OK. `flush=1` definitely doesn't do what we need.

camfindlay:
Sorry misunderstood what you mean. This is expected behaviour of docsviewer when you're told it there should be markdown source and there is none. YOu need to run the build task , get the md source and that will sort of that issue. It could be handled more elegantly in docsviewer when it cannot find a referenced source folder

app/code/RefreshMarkdownTask.php
```diff
@@ -106,6 +106,10 @@ private function cloneRepository(array $repository)
      */
     private function cleanRepository(array $repository)
     {
+	
+	// if the dev=1 flag is used when invoking RefreshMarkdownTask the git repos are kept.
+	if( isset($_REQUEST['dev']) && $_REQUEST['dev'] ) return;
```
camfindlay: Probably needs to change the git clone command too to ensure you get the full history not just the shallow clone.

app/_config/toolbar.yml
```diff
@@ -13,7 +13,7 @@ Only:
   environment: dev
 ---
 GlobalNav:
-  hostname: '//ssorg.dev/'
+  hostname: '//www.silverstripe.org/'
```
camfindlay: thanks

.htaccess
```diff
@@ -29,8 +29,8 @@ ErrorDocument 500 /assets/error-500.html
 <IfModule mod_rewrite.c>
 	RewriteEngine On
 
-	RewriteCond %{HTTP_HOST} ^docs?.silverstripe.com$ [NC]
-    RewriteRule ^(.*)$ https://docs.silverstripe.org/$1 [L,R=301]
+	RewriteCond %{HTTP_HOST} ^docs?.silverstripe.org$ [NC]
```
camfindlay: Please change this back

SpiritLevel:
But, the error occurs while running the build task `RefreshMarkdownTask` so running it doesn't fix the problem. I had to manually add the missing directories to get it to run. Perhaps we should include in `RefreshMarkdownTask` something to create the needed directories if not present or remove the code that checks for their presence.

app/code/RefreshMarkdownTask.php
SpiritLevel: Darn...there goes some of the efficiency gain you made :(

.htaccess
SpiritLevel: Sorry...not quite understanding....what do you want it changed back to ?

camfindlay: the condition is if request comes in on our .com domain eg docs.silverstripe.com it redirects to the correct docs.silverstripe.org domain. What you have here will create an infinite loop when you get it on the actual production server.

SpiritLevel: Ahh...I see what you are are wanting .....your right about the infinite redirects....what I had was ok when the "s?" was removed....but now it loops...oopsie.

So, you want the .com and the "s?" back in ?

camfindlay: Put it as it was when this code was merged in to master ;)

app/code/RefreshMarkdownTask.php
SpiritLevel: Done.

SpiritLevel:
@camfindlay OK, when `RefreshMarkdownTask` is run, it is run within `doc.silverstripe.org` and hence `docsviewer` first gets fired up, which immediately reads `docsviewer.yml`. Then, upon constructing a `DocumentationManifest`, the constructor checks to see if the files registered in `docsviewer.yml` are actually present. If they aren't, an error is thrown, which is what happens running the task with a fresh clone and composer install of  `doc.silverstripe.org`. This occurs even before `docs-repositories.yml` is read.

This cannot be fixed easily by simply creating the needed directories in `RefreshMarkdownTask` since `DocumentationManifest` is run first. 

The easiest fix for this directory problem is to just create empty directories `assets/src/framework_master` etc. in the github repo `doc.silverstripe.org` so that they exist after the initial cloning. I'll implement this fix now but there must be a better solution.

Even with the above solution to the directory problem, the uncomfortable duplication of config information would remain. I'm not sure of the best way to handle this but it definitely needs fixing: one only wants a **single** place where the documentation files and versions are registered. This seems to suggest additional reorganisation of `docsviewer` is needed....

Ideas/suggestions anyone ? @dhensby ?

SpiritLevel:
@camfindlay I noticed the site has reverted. I am on standby to do more once sorted :)

camfindlay:
@SpiritLevel and... we're back up on the new servers. If you want to continue. There is also some noise over on the docsviewer module you might want to check in on.

SpiritLevel:
@camfindlay Ok. Where do you want to go from here ? Should I separated these into separate PRs ? See also https://github.com/silverstripe/silverstripe-cms/issues/1407

camfindlay:
Separate PRs would be useful and easier to test

SpiritLevel:
OK, will separate...

PR #117, opened 2016-02-18
Make the build task more efficient and robust

camfindlay:


camfindlay:
@assertchris peer plz

app/code/RefreshMarkdownTask.php
```diff
@@ -94,11 +94,9 @@ private function cloneRepository(array $repository)
         $this->printLine("cloning " . $remote . "/" . $branch);
 
         chdir("{$path}/src");
-        exec("git clone -q git://github.com/{$remote}.git {$folder}_{$branch} --quiet");
+        exec("git clone -q https://github.com/{$remote}.git {$folder}_{$branch} --depth 1 --branch {$branch} --single-branch");
```
SpiritLevel: Nice :) At one stage, I did look to see if you could only clone one directory (ie: docs) but that isn't possible.

assertchris: Might still be able to... https://github.com/silverstripe/doc.silverstripe.org/issues/118

PR #123, opened 2016-02-29
Update to feedback module + core update

camfindlay:


camfindlay:
@mandrew updated of your module on docs site (leaving core as is).

Issue #118, opened 2016-02-18
Potential for cloning a sparse subfolder of each source...


assertchris:
`git archive --remote=<repo_url> <branch> <path> | tar xvf -`

I'm not sure if this works with the `--depth`, `--branch`, and `--single-branch` flags of `git clone`.

SpiritLevel:
Unless I read something incorrectly, it looks like `git archive` wouldn't keep the git history and hence wouldn't be any good for those that wanted to contribute (like me!).

camfindlay:
Hence the mention of some sort of "dev" flag in those tasks

SpiritLevel:
OK...I added a dev flag :)

SpiritLevel:
See https://github.com/silverstripe/doc.silverstripe.org/pull/129

camfindlay:
Perhaps have a crack at the sparse cloning to gain even more speedy task run :D

SpiritLevel:
@camfindlay 

GitHub doesn't allow the use of `git archive`: https://twitter.com/GitHubHelp/status/322818593748303873

However, if subversion is running on docs.silverstripe.org, we could use this:

`svn export https://github.com/silverstripe/silverstripe-framework/branches/3.3/docs /local/folder/docs`

which only copies the docs folder and all of its contents from GitHub :) This would eliminating having to delete non-documentation files after a shallow clone. It works beautifully :)

There is no such thing as sparse cloning, only sparse checkout for which you still have to get the repository. We can't use git archive. Partial cloning by repo or shallow clones  are options but still bring down the repo and require deletion of files (currently we use shallow). See `http://www.peterlundgren.com/blog/on-gits-shortcomings/`. 

I'll submit a PR with the `svn export` command and you can decide if you like or not :)

robbieaverill:
Triaging as an enhancement, we still have problems with how docsviewer downloads the source code (specifically how it deals with it on an NFS share)

unclecheese:
Closing as resolved in rewrite of docs site. Can't say that this issue specifically is resolved, but the builds are faster, and Gatsby caches the cloned respositories.

PR #129, opened 2016-03-22
add dev flag to keep full git repositories to enable local development

SpiritLevel:
whitespace

more whitespace

This has been tested with and without the dev=1 flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,88 +1,89 @@
 # doc.silverstripe.org
 
-This is the source code powering http://docs.silverstripe.org. It
-primarily consists of the SilverStripe
+This repository contains the source code powering SilverStripe's developer documentation website https://docs.silverstripe.org.
+
+The source code here consists primarily of the SilverStripe
 [framework](https://github.com/silverstripe/silverstripe-framework)
 and the [docsviewer](https://github.com/silverstripe/silverstripe-docsviewer)
 module with minimal configuration.
 
-For adding functionality or editing the style of the documentation see the 
-[docsviewer](http://github.com/silverstripe/silverstripe-docsviewer) module.
+**This repository does not contain the most current documentation files**.
 
-## Development
+The most current documentation files are not stored here but instead are
+stored in each framework repository within a `docs` folder. For
+example, the documentation for the master branch of the SilverStripe
+framework is stored in
+[https://github.com/silverstripe/silverstripe-framework/tree/master/docs](https://github.com/silverstripe/silverstripe-framework/tree/master/docs).
+As described below in the Installation section, one must first obtain
+and re-index fresh versions of the documentation files from the framework
+repositories before being able to view the latest content.
 
-To set up a test instance:
+For adding functionality or editing the style of the documentation, see the 
+[docsviewer](https://github.com/silverstripe/silverstripe-docsviewer) module.
 
- * Clone this repository to a LAMP server.
- * Install [Composer](http://docs.silverstripe.org/en/getting_started/composer)
- * Install [sake](https://docs.silverstripe.org/en/developer_guides/cli/).
- * After installing composer run `composer install --prefer-source` to grab the modules.
- * Run the docs crontask in the browser `dev/tasks/UpdateDocsCronTask`
-   to download all fresh markdown documentation files and reindex them. Note: this
-   will take some time to run. Alternatively, you can use sake
-   to perform these tasks by firstly running the command `sake
-   dev/tasks/RefreshMarkdownTask flush=1` and secondly `sake
-   dev/tasks/RebuildLuceneDocsIndex flush=1`.
- * Make sure to flush the cache for markdown content to show up.
+## Installation
 
-## Source Documentation Files
+To set up a local instance of `doc.silverstripe.org`:
 
-Documentation for each module is stored on the filesystem via a full git clone
-of the module to the `src/` subdirectory in this project. These checkouts are
-ignored from this repository to allow for easier updating and to keep this
-project small.
-
-To update or download the source documentation at any time run the following
-BuildTask command with sake:
-
-	cd /Sites/doc.silverstripe.org/
-	sake dev/tasks/RefreshMarkdownTask flush=1
-
-This build task will download / update each module as listed in the
-`app/_config/docs-repositories.yml` file. Running `sake
-dev/tasks/RebuildLuceneDocsIndex flush=1` will then create a search
-index and reindex the documentation to facilitate searching.
-
-Once the build task has executed and downloaded the latest files,
-those files are registered along with the module version the folder relates to
-through the `app/_config/docsviewer.yml` file.
-
-```yaml
-DocumentationManifest:
-  register_entities:
-    -
-      Path: "src/framework_3.2/docs/"
-      Title: "Developer Documentation"
-      Version: "3.2"
-      Stable: true
-      DefaultEntity: true
+* Install [Composer](https://docs.silverstripe.org/en/getting_started/composer).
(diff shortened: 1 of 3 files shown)


dhensby:
@assertchris thoughts?

SpiritLevel:
I have also locally updated the README but I'll wait until @camfindlay and @assertchris comment before submitting PR.

SpiritLevel:
@assertchris Perhaps one could use the archive idea you had in https://github.com/silverstripe/doc.silverstripe.org/issues/118 rather than the shallow cloning I used in this PR (when the dev=1 flag is NOT used). Archiving might be faster than the shallow clone and may also eliminate the need for deleting non-markdown files, improving performance a bit more.

camfindlay:
@SpiritLevel this looks good, I'll check out and test it at my end and merge once I've cited it working :+1:

app/code/RefreshMarkdownTask.php
```diff
@@ -78,40 +76,43 @@ private function getRepositories()
...
-        $paths = array_merge(glob("*"), glob(".*"));
+        // If the dev=1 flag is used when RefreshMarkdownTask is run, a full git clone of the framework repository is kept
+        // to enable local development of framework and doc.silverstripe.org from within doc.silverstripe.org. Otherwise,
+        // only a shallow clone of the framework repository is made and all non-markdown files deleted, only allowing viewing
+        // of documentation files.
 
-        foreach ($paths as $path) {
-            if ($path !== "docs" && $path !== "." && $path !== "..") {
-                exec("rm -rf {$path}");
+        if( isset($_REQUEST['dev']) && $_REQUEST['dev'] ) {
```
camfindlay: @SpiritLevel can you use the correct SilverStripe framework reference to these variables. i.e. $this->request->getVar('dev');

app/code/RefreshMarkdownTask.php
```diff
@@ -30,13 +30,12 @@ class RefreshMarkdownTask extends BuildTask
      */
     public function run($request)
     {
```
camfindlay: Add a class scope variable $this->request = $request; to capture the passed in request and use it later on in the class.

camfindlay:
Couple of changes to make @SpiritLevel - also in the `UpdateDocsCronTask`, add `$this->request = Controller::curr()->getRequest();` around line 18 and pass into the calls to `run()` instead of `null`.

Other than that it works well, just keen to tidy up and tap into framework vs. raw php $_REQUEST

SpiritLevel:
OK, changes made :) Please check. When merged, I'll update README.

BTW, I was just emulating the approach used here with the parameter:

https://github.com/silverstripe/silverstripe-docsviewer/blob/master/code/tasks/RebuildLuceneDocsIndex.php#L40

camfindlay:
Seems fine, update the README as part of this PR, also add the following to the other CronTask then I think we're good to merge:

```php
// in UpdateDocsCronTask run() method
$this->request = Controller::curr()->getRequest();
//replace the null in the two arguements e.g. $refresh_task->run(null);
$refresh_task->run($this->request);
```

app/code/UpdateDocsCronTask.php
```diff
@@ -3,11 +3,16 @@
...
+    */
+    protected $request = null;
+
+    /**
      * @return string
      */
     public function getSchedule() 
     {
-	return "0 20 * * *";
+	return "0 22 * * *";
```
camfindlay: ?

SpiritLevel: Too fast! I changed it to 22 temporarily to test locally since it is 22 where I am :)

SpiritLevel:
OK updated README and UpdateDocsCronTask :) Note that I did not include in README the commands for running the tasks directly in the browser, only by sake. The reason is that in sake one sees the progress of the tasks whereas in the browser the tasks appear to complete instantly, which is misleading.

README.md
```diff
@@ -1,88 +1,87 @@
...
- * After installing composer run `composer install --prefer-source` to grab the modules.
- * Run the docs crontask in the browser `dev/tasks/UpdateDocsCronTask`
-   to download all fresh markdown documentation files and reindex them. Note: this
-   will take some time to run. Alternatively, you can use sake
-   to perform these tasks by firstly running the command `sake
-   dev/tasks/RefreshMarkdownTask flush=1` and secondly `sake
-   dev/tasks/RebuildLuceneDocsIndex flush=1`.
- * Make sure to flush the cache for markdown content to show up.
-
-## Source Documentation Files
```
camfindlay: We need to keep this bit @SpiritLevel so others know how to add new documentation source / versions in the future.

camfindlay:
We'll need to keep the browser based running of tasks as we dont have sake access on stack share hosting (yet).

SpiritLevel:
@camfindlay How to contribute to or add a new version of docs is now in the Contribution section. 

In the browser, RefreshMarkdownTask appears to execute instantly. Also, I couldn't get RebuildLuceneDocsIndex to run directly in the browser. If you can run these in the browser, perhaps you could quickly add the needed lines to README ? I am only comfortable writing in the README what works for me locally :)

README.md
SpiritLevel: See Contribution section.

camfindlay:
Ok, I'll deal with the browser stuff. We still need to ensure that the information about which yaml config files need to be updated if a future developer needs to include new versions of source documentation. This I think is still missing in this edit from the original readme doc.

README.md
```diff
@@ -1,88 +1,87 @@
...
-project small.
-
-To update or download the source documentation at any time run the following
-BuildTask command with sake:
-
-	cd /Sites/doc.silverstripe.org/
-	sake dev/tasks/RefreshMarkdownTask flush=1
-
-This build task will download / update each module as listed in the
-`app/_config/docs-repositories.yml` file. Running `sake
```
camfindlay: We need to include this in docs.

README.md
```diff
@@ -1,88 +1,87 @@
...
-dev/tasks/RebuildLuceneDocsIndex flush=1` will then create a search
-index and reindex the documentation to facilitate searching.
-
-Once the build task has executed and downloaded the latest files,
-those files are registered along with the module version the folder relates to
-through the `app/_config/docsviewer.yml` file.
```
camfindlay: We need to include this in docs and below code example to help others add source later.

README.md
```diff
@@ -1,88 +1,87 @@
...
-
-```yaml
-DocumentationManifest:
-  register_entities:
-    -
-      Path: "src/framework_3.2/docs/"
-      Title: "Developer Documentation"
-      Version: "3.2"
-      Stable: true
```
camfindlay: Keep this code example, perhaps mention that only one version should be set to stable. Yes, this is in the docsviewer module docs however since this is not a module, rather a project and adding versions is a regular thing we should document this here.

README.md
SpiritLevel: See the first section lines 3-19 for an explanation of how documentation source files are stored and accessed. 

See Contribution lines 81-84 for an explanation of which YAML files need updating.

See Installation lines 40-53 for the the sake commands used to run the tasks.

README.md
SpiritLevel: I think the code example is superfluous since the syntax is in the YAML file itself. Why duplicate ?

I'll add the comment about setting only one version to stable at a time.

README.md
SpiritLevel: See above comment for where all this information is now in README.

I think if you read README in its entirety, rather than the diffs here, you'll see that all the bases are covered....:)

Issue #13, opened 2012-10-03
Expand panels from start


wolfv:
It would be much quicker for me if all panels below "In this module:" are expanded from start, because I never know what is below Reference, Misc or Topics. 

In that case I could Strg+F the page, because the search in silverstripe is *horrible* slow... :)

Issue #10, opened 2012-09-21
Changelogs?


wilr:
This was brought up by @halkyon and me having a chat about the changelog docs being out of date and it got me thinking, what's the reason we have the changelogs on the wiki? They don't _actually_ get updated ever after they're created.. We do a number of things to reduce the changelogs being shown in the search and it can get annoying doing a global search in my editor to sift through the changelogs.

Perhaps a better way would be to publish the changelogs to silverstripe.org/changelogs and simply leave the documentation for upgrade 'notes' on each version and how tos rather than just a git log.

Thoughts?

PR #131, opened 2016-03-29
switched to svn export command for copying documentation files

SpiritLevel:


camfindlay:
This assumes svn is available on SilverStripe Platform (pretty sure it's not).

SpiritLevel:
Since svn is called in an exec command, isn't it only required that svn be on docs.silverstripe.org ?

SpiritLevel:
As long as svn provides some tools that git doesn't (see http://www.peterlundgren.com/blog/on-gits-shortcomings/ for examples), might it be a good idea to incorporate use of svn into SilverStripe Platform? This PR is a case in point; there is no git command that does what needs to be done at 8PM every night on your docs server!!!  Platform allows `exec` and `chdir` etc. so I imagine installing subversion and making it available via `exec` can't be too difficult ?

camfindlay:
Platform is standardised, software configured stacks so unlikely for now that svn will be getting in there. Not to discount the idea though, interesting and seems like a fast approach :+1: I'm going to close for now, let's get back onto the rewrites and such :)

SpiritLevel:
@camfindlay Hiya...I'd keep this open as a reminder :) 

Have you asked anyone at SS if svn can be made available in Platform ? Since svn is still reasonably popular, wouldn't it be a commercially sensible thing to do to make it available in Platform ? @hafriedlander @sminnee @chillu @tractorcow ?

To be honest, your explanation of why svn isn't available in Platform evokes skepticism in me, but that doesn't take much effort, being skeptical by nature. And, my complete ignorance of Platform helps fuel the skepticism too! :P 

From a totally naive point of view, if a program like git is accessible in Platform, how hard can it be to make svn accessible too ? It doesn't have to be deeply integrated into Platform, like I am assuming git may be. Or it could be not integrated at all: surely, at a minimum and as a one-off, you should be able to just install subversion on the machine hosting the docs site and then then use `exec(svn export ... )` from within the docs site. Or, is 'standardised ' a euphemism for 'handcuffed' ?!? :P

dhensby:
I get that we should probably not be tightly coupling to git here, but the effort to loosely couple is probably too high given we're the only ones hosting it and SVN will never be available on our hosting environment

SpiritLevel:
**Re: git, svn and Platform. **

Hey @dhensby. Note that I wasn't saying that git should NOT be tightly coupled/integrated into Platform; that seems reasonable since git is a very popular and very stable bit of software.  I was only saying the svn should be made available for those small number of version control processes that git just can't do. And, svn could be made available in a way that is not coupled at all: just make it present so that developers can use it in those rare cases where git isn't enough. Again, this PR is a perfect case in point: without svn on the docs site, every day at 8PM the framework repos are shallow cloned and then most of the files subsequently deleted. I imagine that it will be a reasonably frequent occurrence that one will want to download bare files from git repos but not the repos themselves. While this is not too wasteful for the framework repos, it could easily get very wasteful if the size of the download is larger.  Maybe someone could give a clearer explanation as to why it isn't easy to make svn accessible on the docs site or why "SVN will never be available on our hosting environment" ? I am curious what the answers would be :)

**Re: rewrites and such:**

Howdy @camfindlay @chillu I am not sure how to approach the rewrites any more as result of various comments I've read. :( So, I had a look a how Laravel organises its docs. It appears that Laravel is doing something similar to what was being done here: some kind of partial mapping between different versions of the docs was set up so that someone reading about, say, template syntax in one version can just click on another a different and be mapped to the corresponding template syntax docs in the new version selected. As proof that Laravel's mapping is only just partial, go to                        

https://laravel.com/docs/4.2/introduction

and then try switch to any other version. You'll get a message about a psychologically unstable housekeeper! And this is on a simple aspect of the documentation, the introduction!!!! There are quite a few other examples like this on the Laravel site.

As I see it, one priority is to drastically reduce the complexity of the set of rewrites rule in the .htaccess file, and, another is to make the process of maintaining them as easy as possible. Here are the options:

**(1)** Abandon any attempt at mapping between versions and force the docs user to go to the top node of the document tree whenever they change versions. The advantages of this approach are that there would essentially be no rewrites to maintain, and, any future structure of the documentation is not at all limited by having to be mappable to and from any prior version, ie, the versions would be "uncoupled". 
The ONLY disadvantage I can see is having to go to the top of the tree when changing versions. 

**(2)** Use a complete mapping between versions. The ONLY advantage is being able to move from one version to another without having to go via the top node of the tree. The disadvantages are having to maintain and update a complex set of rewrite rules for each minor version, and having to accept restrictions on the structure of any new documentation so that one can define a mapping between new and existing documentation without too much trouble.

**(3)** Use a partial mapping between versions. This is what Laravel is doing and what SS is currently doing. Advantages: most of the time, one can navigate between versions without going via the top node, the set of rewrite rules to maintain is slightly less complex than in the case of a complete mapping, and there is some freedom in how documentation can be structured. Disadvantages: the set of rewrite rules to maintain would still be complex, there will be cases in which the user will have to change versions via the top node,  there will still be restrictions on the structure of future versions of the docs, and, there will be the added complexity of deciding upon and acceptable level of incompleteness of the mapping.

In my mind, I was aiming for (1). I was thinking that the standard use case is that of a developer who is interested in reading the docs to help with site development **within a particular version of SS**. Such a developer is unlikely to be switching between versions all that often. 

For contributors to SS, I can see that being able to easily compare a particular topic across different versions of the docs would be useful, and hence (2) or (3) would be of interest to them. However, there are alternatives to this, such as the changelogs and the git repos. The changelogs record what has changed between adjacent versions. And, git is even more flexible: when you are reading documentation markdown files in one version, you can just quickly checkout a different branch to see it in another version, although, again, there is the still issue of partial mapping : if you are in some folder and checkout a branch that does not have this folder then you are forced back up the folder structure.

So, some questions/comments for further discussion:

**(i)** With changelogs and with git repos already available, is it really necessary to have a third route (the dev docs as the main candidate for this route) for switching between docs versions without going via the top node ? 

**(ii)** If yes to (i), then it needs to be decided whether or not (2) or (3) is the best option. And then, some thought needs to go into the nature of the mapping and what restrictions on the tree structure are acceptable. For example, it is likely that all docs versions for all time will have an "Installation" section so this will be easy to set up mappings for. But, within "Installation", documentation folders pages may come and go. Do you allow this level of flexibility ? If so, then you have to maintain complicated rewrite rules for them. If not, then you are imposing restrictions on what can be within the "Installation" section and are hence limiting how the documentation can evolve from minor version to minor version. 

**(iii)** If no to (i), I propose we just go with (1) and make life simpler and easier. And, to better handle 404s, we could redirect these requests to a useful page in the docs (automatically generated index or detailed table of contents) rather than an uninformative 404 page. Also, we could set up a really light weight markdown viewer (ie: a stripped down version of docsviewer) whose only functionality is to display markdown in the same style as the dev docs but completely ignore folder structure. Instead, one could have a file browser for selecting which markdown file to view. Contributors could then use this on a file by file basis to write and compare markdown files.

I am happy to continue cleaning up the links within the docs and I am happy to comb through and sort out the rewrite rules and I am happy to poke around in docsviewer, all as my way of contributing to SS for the free use of the great, open source software SS is. But, I have suspended work on this until there is clear agreement on what kind of mapping for documentation should be put in place :) Over to you!

David.

camfindlay:
Thanks for the detailed comments @SpiritLevel - I'll digest and provide some direction. 

Key for me is to not lose sight of the value of visitors trying to use the docs site not hitting 404s from external inbound links from SERPs etc. At a high level, that's what this is about 👍  I'd like to improve that and tackle the mapping of versions pages as a separate thing once we reduce the amount of 404 errors.

SpiritLevel:
Cool :) 

In dealing with 404s, it seems sensible to deal with self-generated 404s **first**, which is what the above is about. After fixing these and leaving a bit of time for the bots to do their thing, we can then start monitoring 404s, which primarily ought to be from externally constructed URLs that are incorrect. 

Some of the incorrectly constructed URLs are just plain garbage. For example, the CSV file of 404s you sent me had quite a few of these. Many are likely to be internally generated (ie: from a combination of incorrectly hard coded links as well as the dodgy parsing that was going on in docsviewer as it processed api shortcode links). However, many are also likely to be garbage URLs that are externally constructed. I don't think there is anything we can do to accommodate requests based on externally generated 404s with garbage URLs but we can deal with the internally generated ones. We can also deal with internally generated 404s that are not from garbage URLs.

I think we need to clean up all internally generated 404s **before** doing anything else. Only until after an internal clean up will you be confident that the non-garbagey URLs generating 404s are the ones you want to redirect to accommodate visitors. Otherwise, you'd be trying to decide on reasonable redirects while coping with all the 404 noise from both internally and externally generated garbage URLs **AND** the 404 noise from internally generated non-garbagey URLs. That would just be headache inducing and would lead to, well, the kind of .htaccess file you currently have!! It would be far better to eliminate all the internally generated noise first, so that the only noise left over is that due to externally constructed garbagey URLs.

sminnee:
Setting aside the question of whether the `svn` binary is available (if it's something important, we can probably resolve it) I'm a little confused as to what the benefit of this change is? I guess it avoids having to download the non-doc part of modules? Was the old code broken?

Then there's a big discussion about 404s and rewrites. I can't quite see how the git/svn thing relates to the 404s. Am I missing something?

SpiritLevel:
Hi @sminnee Sorry for the verbosity! Just trying to be understood :) 

Short answers: 

The old shell script for updating the docs site wasn't broken but a new method had to be set up when docs was switched over to Platform. The benefit of this change is a reduction in the size of the download involved in the daily updating of markdown files on the docs site. There is no _direct_ relation between this git/svn/Platform PR and the rewrites, so you've not missed anything!

Longer answers: 

The overall context of this work is a desire to reduce 404s and clean up the .htaccess file for docs: I started with the 404s and rewrites on docs. I quickly identified that docsviewer was incorrectly converting api shortcode links into malformed URLs and hence was an important source of 404s. I fixed the api link parser as well as many other links in the docs from 2.4 up to 3.2. While doing this, 
docs was switched over to Platform and the shell script in docsviewer for updating the markdown files had to be replaced by a BuildTask. While this BuildTask task was taking shape, suggestions were made on how to avoid downloading the entire repo for each framework and then subsequently deleting most of it. This is the indirect link between 404s and this PR! 

The first option for reducing the download size implemented in the new BuildTask used a git shallow clone. But, this still involved wasteful downloading and deleting most of what was downloaded. Moreover, by using a shallow clone and then deleting most of the repo data, it made it difficult to contribute. I added a dev flag to the BuildTask which enables one to keep the full repo for contributing. But, this still left the daily docs site updates using the shallow clone, and hence still left the issue of how to get only the markdown files for the daily update of docs. 

While investigating what cloning options git offers to reduce the amount downloaded, it was discovered that the `git archive` command might be useful. However, it was also discovered that GitHub doesn't allow the use of this command and aren't likely to! Further investigation revealed that `svn export` did exactly what we wanted and svn is fully supported by GitHub. This PR has the dev flag to determine if the full repos are kept for contribution, or, if a simple `svn export` download of specific markdown files is to be performed. This solution appears optimal to me and I like optimal solutions.

So, the benefit of the changes are that contributors can get all the repo data (as in the old shell script) by using the dev flag but now the regular updates to the docs site only involve a minimal downloading of markdown files.

The big discussion about 404s is an attempt to get agreement on how to reduce the 404s and clean up and simplify the maintenance of the .htaccess file for the docs site, while at the same time providing useful redirects to visitors. We have been chipping away at this for many months.

Lastly, about making svn available in docs. I am a minimalist and hence loathe waste. The current method of updating the docs site is somewhat wasteful and hence deeply disturbs my sensitivities. :P If it isn't too much trouble, I'd be tickled pink if we could get svn running on the docs site so this PR could get merged in, at a minimum. But it also seems to me to be commercially valuable to allow customer sites managed by Platform to call svn if needed. Some Platform customers may want to use svn on occasion. Is this relatively easy to do ?

sminnee:
@dhensby @camfindlay we can add the subversion Debian package to the relevant platform stack. It'd be this version: https://packages.debian.org/jessie/subversion

I'll leave it to you guys to review the suitability of the change.

sminnee:
@SpiritLevel thanks for filling me in on the context. :-)  We try not to "kitchen sink" the Platform servers as every package you add is a potential support burden, security risk, or performance degradation. However, subversion is a pretty safe choice and so I've added a request to get it added as an optional package on this specific instance. We'd probably hold off putting it into the standard Platform config until after we saw a substantial number of our customers asking for it.

camfindlay:
@SpiritLevel can you add some docs to this to inform future devs they might need to install svn locally if contributing to this project 👍  I'd also be interested in benchmark of the time saved (just coz data is fun).

SpiritLevel:
@sminnee Yay! Thanks! Getting svn on docs, supplemented by a little Zen meditation, and I think my inner peace will be restored soon :P 

I fully understand the "avoid the kitchtchen sink" approach, being a minimalist myself...the less that is there, the less that can go wrong, the less you have to read and hold in your mind, and the less you have to debug. 

It will be interesting to see if additional requests for svn on Platform arise.

@camfindlay Sure can :) I am in Edmonton now so I can only compare svn export to shallow clone from this location. Hopefully back in beautiful NZ soon and will repeat comparison from there. For a proper comparison, we have to include both the shallow clone time and the delete time and compare this against the pure download time of svn export.

camfindlay:
@SpiritLevel agreed on the compare. FYI, the Stack this is on is in the States though since we are comparing locally it's the difference that matters not where it is run.

SpiritLevel:
@camfindlay When I update the docs, do you want to add anything about running `RefreshMarskdownTask`, `RebuildLuceneDocsIndex`, and `UpdateDocsCronTask` in the browser ? Currently, I have only put in how to run then with `sake`.

camfindlay:
You could just mention "run these tasks either with sake or via /dev/tests/ in the browser"

SpiritLevel:
@camfindlay OK, I updated the README :)  Perhaps you could look at the Contribution section. The wording is a little strange and I thought maybe you could think of better phrasing....

SpiritLevel:
@camfindlay @sminnee  Here are the execution times for full git clones, shallow git clones, and svn export of the documentation files:

**The old shell script method that gets full git repos:**

Full git cloning averages **160.68** seconds, which is **16.3** hours per year of download time to SS docs server.

The dev flag now invokes this option and this was the method used on SS docs until now.

**For the non-development option:**

Shallow git cloning average **53.5** seconds, which is **5.4** hours per year of download time to SS docs. server.

Using svn exporting, the average is **23.8** seconds and a yearly SS docs download time of **2.4** hours.

So, svn export gives an 85% time decrease from the original method and a 55% time decrease from the shallow git clone. 

**Data**

10 Trials of RefreshMarkdownTask using full git clones
=========================================
Trial 1 : 129.3821349144 seconds
Trial 2 : 268.94770693779 seconds
Trial 3 : 146.09239220619 seconds
Trial 4 : 131.1084420681 seconds
Trial 5 : 126.34681797028 seconds
Trial 6 : 128.26873803139 seconds
Trial 7 : 137.56154489517 seconds
Trial 8 : 138.13678789139 seconds
Trial 9 : 264.44038009644 seconds
Trial 10 : 136.55593085289 seconds
=========================================
Average : 160.6840875864 seconds
=========================================

10 Trials of RefreshMarkdownTask using git shallow clone
============================================
Trial 1 : 47.3146879673 seconds
Trial 2 : 48.727844953537 seconds
Trial 3 : 48.308659076691 seconds
Trial 4 : 49.087505102158 seconds
Trial 5 : 48.183317899704 seconds
Trial 6 : 83.436289072037 seconds
Trial 7 : 47.344834804535 seconds
Trial 8 : 51.800488948822 seconds
Trial 9 : 64.058090925217 seconds
Trial 10 : 46.917459964752 seconds
============================================
Average : 53.517917871475 seconds
============================================

10 Trials of RefreshMarkdownTask using svn export
=======================================
Trial 1 : 21.363517999649 seconds
Trial 2 : 30.304445981979 seconds
Trial 3 : 21.388597011566 seconds
Trial 4 : 21.76700091362 seconds
Trial 5 : 21.376767158508 seconds
Trial 6 : 21.46965098381 seconds
Trial 7 : 21.998175859451 seconds
Trial 8 : 20.75019288063 seconds
Trial 9 : 28.108836889267 seconds
Trial 10 : 29.760879039764 seconds
=======================================
Average : 23.828806471825 seconds
=======================================

All tests were done on a laptop with a wired ethernet connection to a cable modem from Edmonton.

SpiritLevel:
FYI @camfindlay. This updated README includes a statement about being able to run the tasks in the browser but I tried them and they aren't working.

dhensby:
Leaving this to @camfindlay ;)

camfindlay:
@SpiritLevel just awaiting svn to be rolled out then will look to deploy and test this before merge.

sminnee:
@camfindlay it's rolled out and waiting a full deploy. Click the button at will! :-)

SpiritLevel:
Do it! Click it now!!!! (imagine an Arnold Schwarzeneggar voice here)

Just need to sort out running the tasks in the browser, or, remove comments about this from README...

camfindlay:
Tested on UAT, seems legit... :)  Will look to get some tim e and get the svn rollout on prod servers sorted on Friday.

SpiritLevel:
Excellent :)

dhensby:
well, shall we merge it then?

SpiritLevel:
@dhensby Do you know how to get the tasks running in the browser ? I suppose this should be fixed before merging this (see the README in this PR)...I am a bit busy to investigate and sort this at the moment :(

dhensby:
Go to dev/tasks?

SpiritLevel:
Yip. Tried that for `RefreshMarkdownTask`, `RebuildLuceneDocsIndex`, and tried dev/cron for `UpdateDocsCronTask`

dhensby:
And what's going wrong?

SpiritLevel:
@dhensby :)

OK, I fixed RefreshMarkdownTask by adjusting folder permissions. It now runs in the browser as it should.

RebuildLuceneDocsIndex gives multiple errors and traces. The first trace is:

Trace
exception 'Zend_Search_Lucene_Exception' with message 'Index doesn't exists in the specified directory.' in /home/dra/SpiritLevel/SS/doc.silverstripe.org/docsviewer/thirdparty/Zend/Search/Lucene.php:566 Stack trace: #0 /home/dra/SpiritLevel/SS/doc.silverstripe.org/docsviewer/thirdparty/Zend/Search/Lucene.php(225): Zend_Search_Lucene->__construct('/home/dra/Spiri...', false) #1 /home/dra/SpiritLevel/SS/doc.silverstripe.org/docsviewer/code/tasks/RebuildLuceneDocsIndex.php(49): Zend_Search_Lucene::open('/home/dra/Spiri...') #2 /home/dra/SpiritLevel/SS/doc.silverstripe.org/docsviewer/code/tasks/RebuildLuceneDocsIndex.php(21): RebuildLuceneDocsIndex->rebuildIndexes() #3 /home/dra/SpiritLevel/SS/doc.silverstripe.org/framework/dev/TaskRunner.php(84): RebuildLuceneDocsIndex->run(Object(SS_HTTPRequest)) #4 /home/dra/SpiritLevel/SS/doc.silverstripe.org/framework/control/RequestHandler.php(288): TaskRunner->runTask(Object(SS_HTTPRequest)) #5 /home/dra/SpiritLevel/SS/doc.silverstripe.org/framework/control/Controller.php(202): RequestHandler->handleAction(Object(SS_HTTPRequest), 'runTask') #6 /home/dra/SpiritLevel/SS/doc.silverstripe.org/framework/control/RequestHandler.php(200): Controller->handleAction(Object(SS_HTTPRequest), 'runTask') #7 /home/dra/SpiritLevel/SS/doc.silverstripe.org/framework/control/Controller.php(158): RequestHandler->handleRequest(Object(SS_HTTPRequest), Object(DataModel)) #8 /home/dra/SpiritLevel/SS/doc.silverstripe.org/framework/control/RequestHandler.php(222): Controller->handleRequest(Object(SS_HTTPRequest), Object(DataModel)) #9 /home/dra/SpiritLevel/SS/doc.silverstripe.org/framework/control/Controller.php(158): RequestHandler->handleRequest(Object(SS_HTTPRequest), Object(DataModel)) #10 /home/dra/SpiritLevel/SS/doc.silverstripe.org/framework/control/Director.php(385): Controller->handleRequest(Object(SS_HTTPRequest), Object(DataModel)) #11 /home/dra/SpiritLevel/SS/doc.silverstripe.org/framework/control/Director.php(149): Director::handleRequest(Object(SS_HTTPRequest), Object(Session), Object(DataModel)) #12 /home/dra/SpiritLevel/SS/doc.silverstripe.org/framework/main.php(184): Director::direct('/dev/tasks/Rebu...', Object(DataModel)) #13 {main} 
RebuildLuceneDocsIndex.php:52
RebuildLuceneDocsIndex->rebuildIndexes() 
RebuildLuceneDocsIndex.php:21
RebuildLuceneDocsIndex->run(SS_HTTPRequest) 
TaskRunner.php:84
TaskRunner->runTask(SS_HTTPRequest) 
RequestHandler.php:288
RequestHandler->handleAction(SS_HTTPRequest,runTask) 
Controller.php:202
Controller->handleAction(SS_HTTPRequest,runTask) 
RequestHandler.php:200
RequestHandler->handleRequest(SS_HTTPRequest,DataModel) 
Controller.php:158
Controller->handleRequest(SS_HTTPRequest,DataModel) 
RequestHandler.php:222
RequestHandler->handleRequest(SS_HTTPRequest,DataModel) 
Controller.php:158
Controller->handleRequest(SS_HTTPRequest,DataModel) 
Director.php:385
Director::handleRequest(SS_HTTPRequest,Session,DataModel) 
Director.php:149
Director::direct(/dev/tasks/RebuildLuceneDocsIndex,DataModel) 
main.php:184

And running UpdateDocsCronTask in the browser leads to the standard 404 page for doc.silverstripe.org. I can't tell by running it if the tasks are set to run as cron jobs or not. Will have to investigate later.

camfindlay:
Looks like another permission issue @SpiritLevel I can confirm it works here though.

SpiritLevel:
OK. I'll come back to this after things settle down and I'll work out what needs fixing for `RebuildLuceneDocsIndex` and `UpdateDocsCronTask` for them to work locally in the browser. It all works using `sake`...anyhow, glad its in! :)

PR #132, opened 2016-05-19
Add entry for 3.4 release

tractorcow:


camfindlay:
You'll switch to stable just prior to release I assume? @tractorcow

tractorcow:
Yep, but we can't tag that as stable yet, correct?

PR #133, opened 2016-05-27
Use curl multi-download to update the checked out repositories

stojg:
This will change the normal checkout from a synchronous svn export of the repositories to an asynchronous curl download of tar.gz. I hope this will dramatically download all repos. Current runtime locally is roughly 2 minutes for 7 branches.

Plus contains a fix for running the task via the CLI.

It's a bit involved and a lot of code, but it has a lot more error checking to ensure that it doesn't remove things before the last check is done.

### DIFF
--- a/app/code/RefreshMarkdownTask.php
+++ b/app/code/RefreshMarkdownTask.php
@@ -1,116 +1,307 @@
 <?php
 
+/**
+ * This task will download sources defined in the $documentation_repositories and put them into the
+ * a folder where the markdown parser and indexer can work on them.
+ *
+ * Note that this downloader requires the repositories because it relies on their tarball download
+ * functionality. So this is pretty much hardcoded.
+ *
+ * The tarball downloader runs all the repos in parallel to save time and will not remove the current
+ * source until the new downloaded code can replace the old code.
+ *
+ * For local development you can use dev=1 to instead clone the repositories if you would like to
+ * work on the documentation locally instead.
+ *
+ */
 class RefreshMarkdownTask extends BuildTask
 {
-    /**
-     * @var array
-     * @config documentation_repositories
-     */
-    private static $documentation_repositories;
 
-    /**
-     * @var SS_HTTPRequest $request The request object the controller was called with.
-    */
-    protected $request = null;
+	/**
+	 * @var array
+	 * @config documentation_repositories
+	 */
+	private static $documentation_repositories;
 
-    /**
-     * @var string
-     */
-    protected $title = "RefreshMarkdownTask";
+	/**
+	 * @var SS_HTTPRequest $request The request object the controller was called with.
+	 */
+	protected $request = null;
 
-    /**
-     * @var string
-     */
-    protected $description = "Downloads a fresh version of markdown documentation files from source";
+	/**
+	 * @var string
+	 */
+	protected $title = "RefreshMarkdownTask";
 
-    /**
-     * @var bool
-     */
-    protected $enabled = true;
+	/**
+	 * @var string
+	 */
+	protected $description = "Downloads a fresh version of markdown documentation files from source";
 
-    /**
-     * @var SS_HTTPRequest $request
-     *
-     * @todo test the initial unlink works with cloned modules
-     */
-    public function run($request)
-    {
-        $this->request = $request;
-        $repositories = $this->getRepositories();
-        foreach ($repositories as $repository) {
-            $this->cloneRepository($repository);
-        }
-        $this->printLine(" ");
-        $this->printLine("To re-index the freshly downloaded documentation files either:");
-        $this->printLine("(1) run the command 'sake dev/tasks/RebuildLuceneDocsIndex flush=1' in a shell or");
-        $this->printLine("(2) point your browser at the url 'http://localhost/path/to/ssdocs/dev/tasks/RebuildLuceneDocsIndex?flush=1'");
-    }
+	/**
+	 * @var bool
+	 */
+	protected $enabled = true;
 
-    /**
-     * @return string
-     *
-     * @todo document this new configuration parameter
-     */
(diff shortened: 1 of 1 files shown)


app/code/RefreshMarkdownTask.php
```diff
@@ -1,116 +1,307 @@
 <?php
 
+/**
+ * This task will download sources defined in the $documentation_repositories and put them into the
+ * a folder where the markdown parser and indexer can work on them.
```
SpiritLevel: 'the a' should be 'a' ? :)

app/code/RefreshMarkdownTask.php
```diff
@@ -1,116 +1,307 @@
...
+ * Note that this downloader requires the repositories because it relies on their tarball download
+ * functionality. So this is pretty much hardcoded.
+ *
+ * The tarball downloader runs all the repos in parallel to save time and will not remove the current
+ * source until the new downloaded code can replace the old code.
+ *
+ * For local development you can use dev=1 to instead clone the repositories if you would like to
+ * work on the documentation locally instead.
+ *
+ */
```
SpiritLevel: Once this gets merged, would you mind if I had a go at rewording this ?

app/code/RefreshMarkdownTask.php
```diff
@@ -1,116 +1,307 @@
...
 class RefreshMarkdownTask extends BuildTask
```
SpiritLevel: I'm fussy about names. I changed the name to `RefreshMarkdownTask` but this is not a very good description of what the task does. Sometimes it clones the framework repositories, sometimes it grabs only the markdown files (now as a zipped tarball and checks them). There are a number of other names I'd like to change but I'll wait until this is merged and running.

stojg: Agree, I would probably just make it a repo downloader and reflect that properly in the name.

stojg:
I have seen some regressions with this build that I need to check out, so I'm going to close this for the time being.

app/code/RefreshMarkdownTask.php
stojg: Sure, I would like to reshuffle this a bit as well. I'll give an update after I looked into the regression I've seen on our UAT.

app/code/RefreshMarkdownTask.php
stojg: hah, yes. good catch.

app/code/RefreshMarkdownTask.php
```diff
@@ -1,116 +1,307 @@
...
+		foreach ($repositories as $repository) {
+			list($remote, $folder, $branch) = $repository;
+			$path = $this->getPath();
+			exec("mkdir -p {$path}/src");
+			exec("chmod -R 775 {$path}/src");
+			exec("rm -rf {$path}/src/{$folder}_{$branch}");
+			chdir("{$path}/src");
+			$this->printLine("Cloning repository {$remote}/{$branch} into assets/src/{$folder}_{$branch}");
+			exec("git clone --quiet https://github.com/{$remote}.git {$folder}_{$branch} --branch {$branch}");
+		}
```
SpiritLevel: Are there also potential download problems using `git clone` as there were with `svn` ? Or does `git clone` automatically download to a temp version, check integrity of the temp version, and after integrity verified move the temp version into a final location ? Am I right thinking that looping through the `git clone` commands here is synchronous ? Is there any way to make them asynchronous by submitting them as background jobs, or some other method ?  I know this is only for the development case (as opposed to the regular nightly download of markdown) but I am curious how this might be done :)

app/code/RefreshMarkdownTask.php
```diff
@@ -1,116 +1,307 @@
...
+
+	/**
+	 * Clone $repository which contains the most current documentation source markdown files
+	 *
+	 * @param array $repositories
+	 *
+	 */
+	protected function downloadArchives(array $repositories)
+	{
+		$chs = [];
```
SpiritLevel: Non-descriptive names like this make me a bit crazy! But, it doesn't take much to do that.

app/code/RefreshMarkdownTask.php
```diff
@@ -1,116 +1,307 @@
...
+			$repoInformation[$source] = [
+				'remote' => $remote,
+				'branch' => $branch,
+				'folder' => $folder,
+				'filepath' => $filePath,
+				'status' => null,
+			];
+			curl_multi_add_handle($cmh, $chs[$idx]);
+		}
+
```
SpiritLevel: This is cool!

app/code/RefreshMarkdownTask.php
```diff
@@ -1,116 +1,307 @@
...
+		printf("Downloading %d repositories in parallel\n", count($repositories));
+
```
SpiritLevel: This is really cool :)

SpiritLevel:
Hey! Stig!  What about this:

http://php.net/manual/en/function.pcntl-fork.php

Then maybe you could run multiple `git clone` and `svn export` commands in parallel ?

SpiritLevel:
Here is a bit more "threading" information:

https://d-mueller.de/blog/parallel-processing-in-php/

I am thinking that the first method using '&' to submit jobs to the background might be enough ?

SpiritLevel:
Even better as it looks like my familiar friend from Java:

http://php.net/manual/en/book.pthreads.php

I love using threads...if I had more time at the moment I would dive into this :)

SpiritLevel:
Sorry...last one....it looks like you have some in-house expertise!

https://medium.com/async-php/multi-process-php-94a4e5a4be05#.mde5g6wv1

dhensby:
@stojg is this going to make any progress or shall we delete the branch?

stojg:
yep, thanks for reminding me.

PR #139, opened 2016-06-24
API Rewrite source file update task

tractorcow:
Split updater into a service and provide two default implementations
 - GitMarkdownUpdater
 - SVNMarkdonUpdater

Add logging for errors to SS_Log

Secure escaping of shell arguments and error detection


camfindlay:
Cheers @tractorcow I'll merge and we can get things going on this.

PR #143, opened 2016-10-14
Feature/mobile sidebar nav

camfindlay:
Staging this PR for the fix to the mobile navigation.


camfindlay:
We will await a merge over in docsviewer module before merging this

PR #145, opened 2016-11-04
Hotfix/logo clipping title fix

newleeland:
Added new mobile menu button and sidebar animation to docs site


camfindlay:
@newleeland you'll need to also do a `composer update silverstripe/docsviewer` and commit the new composer.lock file.

camfindlay:
Can you also do the same over on the userhelp site.

PR #154, opened 2016-12-12
Update docsviewer to include versioned "documentation index" page

robbieaverill:
* Add rewrite rules for previous routes
* Add version to matched routing parameters

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -51,6 +51,9 @@ ErrorDocument 500 /assets/error-500.html
 	# Redirect any version > 3.4 to 3 branch
 	RewriteRule ^en/3\.[5-9]/(.*)?$ /en/3/$1 [R=301,L]
 
+	# Redirect old index links to the current version (via another redirect rule)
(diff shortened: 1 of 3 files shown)


.htaccess
```diff
@@ -51,6 +51,9 @@ ErrorDocument 500 /assets/error-500.html
 	# Redirect any version > 3.4 to 3 branch
 	RewriteRule ^en/3\.[5-9]/(.*)?$ /en/3/$1 [R=301,L]
 
+	# Redirect old index links to the current version (via another redirect rule)
```
tractorcow: Can you redirect this to `/en/3/all` instead?

.htaccess
```diff
@@ -51,6 +51,9 @@ ErrorDocument 500 /assets/error-500.html
...
+	RewriteRule ^en/all\/?$ / [R=301,L]
```
tractorcow: Can you redirect this to /en/3/all instead?

PR #167, opened 2017-09-13
NEW Add raygun.io integration module

robbieaverill:


.gitignore
```diff
@@ -12,3 +12,4 @@
 /silverstripe-dynamodb/
 /crontask/
 /quickfeedback/
+raygun/
```
dhensby: should this not be `/raygun/` (with a leading `/`)?

robbieaverill: Yes, good catch. Hotfixed with 554070a

Issue #156, opened 2017-02-01
Avoid outdated docs showing up in search results


chillu:
Google and other search engines are indexing way outdated versions for doc.ss.org - it seems quite arbitrary which version you get on a particular search term.

![image](https://cloud.githubusercontent.com/assets/111025/22523998/4502af52-e926-11e6-80dc-1e283946fa61.png)

We can't redirect from older versions (e.g. `3.0/reference/searchcontext/` to `3.5/reference/searchcontext/`) since that'll make it impossible to get to older versions (which sometimes is still necessary). We could switch some `.htacess` behaviour based on the user agent, with redirects only for search engines, but I'd assume that'd bad SEO juju.

I'd suggest we use [canonical meta tags](https://support.google.com/webmasters/answer/139066?hl=en#2) to solve this. It strictly speaking isn't the exact same content, so might require more research into the SEO impact here.

Note that a separate issue about [rewriting to current](https://github.com/silverstripe/doc.silverstripe.org/issues/119) exists, which is related but won't fix the issue at hand on it's own.

robbieaverill:
@chillu FYI I've just had a quick look at the Symfony docs and it looks like they're pointing canonical tags at the "current" version.

chillu:
I've ensured this pops up in the CC backlog through an internal ticket at https://silverstripe.atlassian.net/browse/SITES-2111

dhensby:
This is a perfectly acceptable use of canonical tags. There's not requirement that a canonical tag points to a version with the exact same content - you're telling search engines where the canonical resource is.

robbieaverill:
Done in #161 and deployed. I didn't add the `/en/current` redirect for now, since `/en` does it already. This could still be done in #119.

chillu:
Legend, thanks Robbie and Franco! :) I've put a reminder in for myself to check if that actually fixes the issue, once Google has reindexed.

PR #180, opened 2018-07-03
FIX: Remove canonical links to try and fix Swiftype indexing.

sminnee:
Swiftype support have suggested that our canonical links might be
preventing correct indexing. As a trial of this, this patch disables
the canonical links.

### DIFF
--- a/themes/docs/templates/Includes/DocumentationHead.ss
+++ b/themes/docs/templates/Includes/DocumentationHead.ss
@@ -6,9 +6,11 @@
 
 	<% include DocumentationFavicons %>
 
+	<%-- disabled for Swiftype indexing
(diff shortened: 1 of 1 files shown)


robbieaverill:
These were put in place to remove duplicate indexing in Google - I'd rather not remove it to allow us to use arbitrary search plugins to be honest. Google is pretty important!

robbieaverill:
Unless of course we've made an incorrect assumption about the fact that we should be using canonical URLs across versions in the first place (https://github.com/silverstripe/doc.silverstripe.org/issues/156)

sminnee:
OK, maybe providing a sitemap.xml is another way of fixing this; I'll enquire with Switftype.

PR #183, opened 2019-02-21
Update framework to 3.7.3 and docsviewer to 2.0 stable

robbieaverill:
Also adds platform configuration for PHP version (7.1 currently) and references this from the readme in order to prevent confusion about which PHP version to use when building this project.

### DIFF
--- a/.platform.yml
+++ b/.platform.yml
@@ -1,0 +1,2 @@
+php_settings:
+  version: 7.1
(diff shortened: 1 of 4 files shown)


robbieaverill:
"platform.yml is ignored for virtual environments" - oh well, at least it indicates that you need to use PHP 7.1 for the project 😆

Issue #179, opened 2018-06-08
Pull SS3 docs from 3.7 branch while retaining `3` URLs


sminnee:
At some stage in the near future, we're likely to delete the `3` branch, once we stop producing new minor releases.

After that point, the documentation hosted at https://docs.silverstripe.org/en/3/ will need to be pulled from the 3.7 branch instead of the 3 branch.

We don't want to break the URLs either.

tractorcow:
At this point, we may as well change from "branch based docs" to "tag based docs".

I would like the "3" docs to point to "latest 3.x tag". That's something we could do now in anticipation of deleting 3 branch, meaning that once we remove it, the docs don't break.

dhensby:
would that mean not publishing the dev docs at all for any branch?

tractorcow:
Yes that's right. Dev docs could still be found on github.

sminnee:
I don't think tag-based docs are a good idea, as it will further slow down the turnaround time on docs updates.

Also it would confusing to have half the docs be branch based and half the docs be tag based.

tractorcow:
Ok fair enough.

unclecheese:
This is a one-liner in the new Gatsby set up. https://github.com/silverstripe/doc.silverstripe.org/pull/202

PR #202, opened 2019-11-19
Update SS3 docs to use 3.7 branch

unclecheese:
See https://github.com/silverstripe/doc.silverstripe.org/issues/179


sminnee:
I expect that this needs to be cherry-picked to the 3.7 branch to fix this

https://github.com/silverstripe/silverstripe-framework/commit/2facc7c80de47a6458a42498806c133803e767e2#diff-9020e0610c38a4cc1e6af1143bd49130

unclecheese:
Good point. Have cherry picked those two commits down.

PR #209, opened 2019-11-20
Add more flags for the CHILDREN tag

dnsl48:
- only
 - includeFolders
 - reverse

Add "unhideSelf" flag for SilverStripeDocument to override parent's "hideChildren" directives.

### DIFF
--- a/src/components/ChildrenOf.tsx
+++ b/src/components/ChildrenOf.tsx
@@ -11,7 +11,7 @@ const createCards = (children: SilverstripeDocument[]): ReactElement[] => {
                 <div className="card shadow-sm">
                     <div className="card-body">
                         <h5 className="card-title">
-                            <span className="theme-icon-holder card-icon-holder mr-2">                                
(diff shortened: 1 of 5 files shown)


src/components/ChildrenOf.tsx
```diff
@@ -38,32 +38,39 @@ const createList = (children: SilverstripeDocument[]): ReactElement[] => {
...
         return null;
     }
-    let children: ReactElement[] = [];
-    if (!folderName && !exclude) {
-        const sourceNodes = currentNode.isIndex ? getChildren(currentNode, false) : getSiblings(currentNode)
-        children = asList ? createList(sourceNodes) : createCards(sourceNodes);
+
+    var nodes: SilverstripeDocument[] = [];
+
+    if (!folderName && !exclude && !only) {
```
Cheddam: Probably worth adding some inline documentation here to describe what each case means and does.

PR #218, opened 2020-01-15
Removed white background on code.

tiller1010:
Some code is unreadable on https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/reactjs_redux_and_graphql


unclecheese:
Yup, good fix that allows us to put code in callout blocks, but my bigger concern is that the callout block is clearly malformed and taking in more content than it's supposed to. Will make a separate PR to framework docs for that.

PR #219, opened 2020-01-16
NEW Add anchor links to headings

Cheddam:
This PR introduces anchor links to headings in the documentation. These are useful for deeply linking to specific sections of a document. I've also included a small CSS hack for pushing the heading below the header when the anchor link is used - better than not having it there, but I'll be hunting around for a better approach to apply later.

Also includes a commit with some tidying of the README file, as it had some old information about documentation sources and was inconsistently referring to bare gatsby commands, whereas we now recommend using the scripts in `composer.json`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,16 +11,14 @@ site generator based on [React](https://reactjs.org). It sources content
 from the [framework](https://github.com/silverstripe/silverstripe-framework)
 repository for each major release.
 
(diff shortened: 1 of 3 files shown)


unclecheese:
Lovely. Nice work. Probably about time to make those `rewrite*.js` files proper JSX.

PR #226, opened 2020-06-30
BUGFIX: Ensure hash links are not treated as relative

unclecheese:
Resolves https://github.com/silverstripe/doc.silverstripe.org/issues/224

PR #228, opened 2020-06-30
BUGFIX: Ensure .md links with hashes are rewritten properly

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-userhelp-content/issues/145

### DIFF
--- a/src/utils/rewriteLink.ts
+++ b/src/utils/rewriteLink.ts
@@ -20,6 +20,7 @@ const relativeLink = (currentNode: SilverstripeDocument, href: string): string =
     return slug;
 };
 
+
 /**
  * Ensure links use the Gatsby <Link /> component. Client-side routing FTW
  * 
@@ -39,7 +40,6 @@ const rewriteLink = (
 
     const currentNode = getCurrentNode();
     const version = getCurrentVersion();
-
     // shorthand api links
     if (href.match(/^api\:/)) {
         const newHref = rewriteAPILink(href);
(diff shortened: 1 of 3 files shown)


src/utils/rewriteLink.ts
```diff
@@ -102,6 +101,22 @@ const rewriteLink = (
                 domToReact(children, parseOptions)
             )
         }
+        const hashMatches = href.match(/(\.md#([A-Za-z0-9_-]+))$/);
```
emteknetnz, line 104: I think you'll need to escape the final `-`
```suggestion
        const hashMatches = href.match(/(\.md#([A-Za-z0-9_\-]+))$/);
```

emteknetnz:
How come this PR doesn't auto-deploy on the CI like this PR does https://github.com/silverstripe/doc.silverstripe.org/pull/226?

src/utils/rewriteLink.ts
unclecheese: I don't believe you do? I mean, it's working in the preview: https://5efaa115098fad000928d5d0--ss-userhelp.netlify.app/en/4/optional_features/content_blocks/

You can see the raw source here: https://github.com/dnadesign/silverstripe-elemental/blame/4/docs/en/userguide/index.md#L6 where the `.md` is in the link and removed on the rendered docs site.

emteknetnz: ok all good then

PR #231, opened 2021-01-14
Fix docker setup

dnsl48:


GuySartorelli:
It's not clear what this is intended to fix - the current setup seems to work fine.

PR #249, opened 2022-11-21
FIX Fix issue with handling png images

GuySartorelli:
Hopefully fixes  https://app.netlify.com/sites/ss-docs/deploys/637ad1699d407f36cd349198

> Failed to process image `/opt/build/repo/.cache/gatsby-source-git/docs--5/en/_images/cms-architecture.png`
> Input file contains unsupported image format

We had to update the ubuntu version in netlify which has updated `libvips` to a version that doesn't support the version of `sharp` we had installed (see https://github.com/lovell/sharp/issues/2062#issuecomment-580189056).
We couldn't just install sharp because the gatsby didn't support a more recent version.

Hopefully this resolves the problem and we don't have to go up major versions of gatsby.


maxime-rainville (approved):
LGTM

PR #253, opened 2023-02-08
ENH Styling for api links contains backticks

emteknetnz:
Ideally would have been done in ss-docs.scss, though I ran into multiple issues when running `yarn build`

Changes the underline color to grey to match

image


GuySartorelli:
This can happen after the beta I think - it won't materially change anything and we really need to get this out.

Issue #259, opened 2023-03-20
Problems with developer docs - irrelevant results (especially when searching non-default docs)


GuySartorelli:
There are several problems with the search in developer docs right now:
1. If you are searching on v3 or v5 docs, it only gives you results for v4 docs (meaning you sometimes don't get relevant results, and you are always redirected to v4 docs which may not be the same as the docs for the version you were looking for)

## Acceptance criteria
- When searching, you only get results for the version you are currently viewing.
- There's doc on how to configure new search for new major releases.
- Create a a follow card for UserHelp

## Note
- We recently upgraded to Algolia docSearch v3. The search option dashboard there as a lot of new options. There could be some magical fix there.
- The following doc may be useful:
    - [Introduce global information as meta tags](https://docsearch.algolia.com/docs/required-configuration/#introduce-global-information-as-meta-tags)
    - [Indexing content for faceting](https://docsearch.algolia.com/docs/record-extractor/#indexing-content-for-faceting)
    - [Avoid duplicates by promoting unicity](https://docsearch.algolia.com/docs/tips/#avoid-duplicates-by-promoting-unicity) (aka use facets to search a specific version of docs)
    - [Filtering your search](https://docsearch.algolia.com/docs/docsearch-v3/#filtering-your-search) (again, using facets)
    - [Migrating from the legacy scraper](https://docsearch.algolia.com/docs/migrating-from-legacy/)

## PRs
- https://github.com/silverstripe/doc.silverstripe.org/pull/262

maxime-rainville:
When we hit stable, we'll probably want to make the 5 branch the default branch. Probably as part of that process, the 5 branch will become what is being indexed by search results.

GuySartorelli:
In that case we'll still have the same problem - it'll just be different branches that exhibit the problem.

maxime-rainville:
I've put in a separate request for the User Help.

maxime-rainville:
Still waiting for Algolia for userhelp. We should proceed with the developer doc deployment because this is a blocker for the CMS5 release.

GuySartorelli:
I've made a new separate card to handle the userhelp changes: https://github.com/silverstripe/doc.silverstripe.org/issues/263

GuySartorelli:
PR merged and dev docs search is working as expected. New card created for userhelp.

PR #262, opened 2023-04-13
ENH Use the new algolia search index and API

GuySartorelli:
## What's in this PR?
1. Update dev environment variables (note: these get exposed in the browser so they're not secret - it's safe to add them here)
2. Add note about how to update algolia crawler when adding a new major release
3. Replace the old v2 docsearch API usage with the new v3 docsearch API usage, using the new shiny react component
  - Note: The old API took an input element ID, which meant we could style the thing however we wanted. The new API takes a container and completely replaces the container _even if you're not using the react component_, so our styling options are limited. There is [a page in algolia docs for styling](https://docsearch.algolia.com/docs/styling) but it has no useful information in it
4. Update styling for the new search component so it behaves nicely with our header and breakpoints

## Action required after merging
After this is merged, we need to update the environment variables in netlify for the main docs. I'm happy to do that - but make sure I'm ready to do so when you merge this so we don't end up with broken search for an extended period.

## Issue
- https://github.com/silverstripe/doc.silverstripe.org/issues/259


src/components/Header.tsx
```diff
@@ -23,14 +23,9 @@ const Header: StatelessComponent<HeaderProps> = ({ handleSidebarToggle }): React
...
-        const otherNode = getNodes().find(n => n.slug === newPath);
-        // This has to be a hard refresh, because the sidebar needs to unmount
-        if (otherNode) {
-          navigate(otherNode.slug);
-        } else {
-          navigate(`/en/${ver}`);
-        }
+        const newPath = getVersionPath(currentNode, ver);
+        // This has to be a hard refresh, because the sidebar and searchbar need to unmount
+        navigate(newPath);
```
GuySartorelli, line 28: This ultimately still does the same thing it was doing before - it's just a lot cleaner.

src/components/Layout.tsx
```diff
@@ -20,18 +20,21 @@ const Layout: StatelessComponent<LayoutProps> = ({ children, pageContext: { slug
   const currentNode = getCurrentNode();
   const versionMessage = getVersionMessage();
 
+  const helmetLinks = [{
+    rel: 'preconnect',
+    href: `https://${process.env.GATSBY_DOCSEARCH_APP_ID}-dsn.algolia.net`,
+    crossOrigin: ''
```
GuySartorelli, line 26: See https://docsearch.algolia.com/docs/DocSearch-v3#performance-optimization

src/components/SearchBox.tsx
```diff
@@ -1,75 +1,51 @@
 import React from 'react'
-import { StatelessComponent, ReactElement, useEffect, useState } from 'react';
-import { navigateTo } from "gatsby-link"
+import { StatelessComponent, ReactElement } from 'react';
+import { navigate } from "gatsby"
```
GuySartorelli, line 3: `navigateTo` and `navigate` do the same thing, but `navigateTo` is deprecated so may as well stop using that while we're at it.

src/components/SearchBox.tsx
```diff
@@ -1,75 +1,51 @@
...
+          navigate: ({ itemUrl }) => navigate(makeUrlRelative(itemUrl)),
+        }}
+        // Overrides the link container for search results so we can use gatsby navigation instead of full page loads
+        hitComponent={({ hit, children }) => {
+          const relativeUrl = makeUrlRelative(hit.url);
+          // We can't use a gatsby <Link> component here - for some reason it makes the links disappear on hover
+          return <a onClick={(e) => handleClick(e, relativeUrl)} href={relativeUrl}>{children}</a>
+        }}
+      />
+    );
```
GuySartorelli, line 48: See https://docsearch.algolia.com/docs/api

emteknetnz (approved):
Code looks fine, self-merge when you're ready

PR #264, opened 2023-05-04
ENH Make 5 the default version

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/694

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,20 +2,26 @@
   from = "https://doc.silverstripe.org/*"
   to = "https://docs.silverstripe.org/:splat"
   force = true
-  status = 301
+  status = 302
 
 [[redirects]]
   from = "/"
-  to = "/en/4/"
+  to = "/en/5/"
+  status = 302
+
+[[redirects]]
+  from = "/en"
+  to = "/en/5/"
+  status = 302
 
 [[redirects]]
   from = "/framework/en/*"
-  to = "/en/4/:splat"
-  status = 301
-  
+  to = "/en/5/:splat"
+  status = 302
+
 [[redirects]]
   from = "/contributing"
-  to = "/en/4/contributing/"
+  to = "/en/5/contributing/"
 
 [[redirects]]
   from = "/en/4/getting_started/installation/common_problems"
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ For **new stable releases**, you will need to do the following
 - Remove the major from the `PRE_RELEASE` array in the `getVersionMessage` function in `src/utils/nodes.ts`
 - Update the `getDefaultVersion` function's return value to the new stable major in `src/utils/nodes.ts`
 - Update redirects in `netlify.toml`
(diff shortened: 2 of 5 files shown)


netlify.toml
```diff
@@ -2,20 +2,20 @@
   from = "https://doc.silverstripe.org/*"
   to = "https://docs.silverstripe.org/:splat"
   force = true
-  status = 301
+  status = 302
```
maxime-rainville, line 5: What's the rational for making this a temporary redirect?

maxime-rainville (commented):
I take it that 4 and 5 are going to get the "Supported" banner?

image

netlify.toml
emteknetnz: So it doesn't get cached e.g. browser cache.

This should have been a 302 to being with. 301's should only be used for things never intended to be changed again in the future. This will changed again in the future when we change to CMS 6.

emteknetnz:
>  take it that 4 and 5 are going to get the "Supported" banner?

Yes

You can view the deploy preview here https://deploy-preview-264--ss-docs.netlify.app

netlify.toml
maxime-rainville: I think we have different understanding of the best practice here. In my mind, a "temporary" redirect is something that will last a few hours or a few days. e.g.: Something is broken on the site and we will send you somewhere else while we work on fixing it. 

Something that will last for two years, is functionally permanent.

If we set it to a 302, won't Google want to keep the version 4 of the doc as the "default" one in search results?

maxime-rainville:
We might want to add a bit more details to the v4 version to say it's stable but not the latest version.

But that can wait.

GuySartorelli:
> We might want to add a bit more details to the v4 version to say it's stable but not the latest version

So the thing that was on the card that got struck out for some reason? 😂
If that's not being done as part of this pr but is intended to be done later, a card should be created for it and added to the backlog.

netlify.toml
emteknetnz: It's more an issue that 301 redirects can get cached at various points where traffic is served, including in browser caches. This means that browsers that previously visited a url will still redirect to the old old location.  302 doesn't get cached so it doesn't have this problem.

I'd expect google is smart enough to handle changing a 301 to a different location, I'd expect it's a fairly common occurrence that people change decide that a 301 should point somewhere else after some time

We also use 302 on api.silverstripe.org for the default redirect - https://github.com/silverstripe/api.silverstripe.org/blob/master/.htaccess#L49

emteknetnz:
New card was created for the struck out stuff - https://github.com/silverstripe/.github/issues/45

emteknetnz:
> In the preview, if you go to /en/ it takes you to v4 docs still.

Have fixed this

GuySartorelli:
Going to https://deploy-preview-264--ss-docs.netlify.app/en/getting_started/composer/ still takes me to the v4 version of the page

maxime-rainville (approved):
Maxime is temporarily happy

Issue #274, opened 2023-10-17
`[CHILDREN]` blocks are rendered in code blocks (but shouldn't be)


GuySartorelli:
![image](https://github.com/silverstripe/doc.silverstripe.org/assets/36352093/03dbd82b-7c16-4db6-b340-2cdd5cd16008)

The above is meant to be `[CHILDREN]` and the like, i.e. not rendering the children lists but being correctly escaped by the code blocks. See https://docs.silverstripe.org/en/5/contributing/documentation/#link-to-children

## PRs
- https://github.com/silverstripe/doc.silverstripe.org/pull/275

Issue #220, opened 2020-01-16
Callout block parsing is fragile


unclecheese:
The `[callout] [/callout]` syntax has a number of issues, particularly when other markdown is nested in it. See: https://github.com/silverstripe/silverstripe-framework/pull/9375

Ideally this should be done with a plugin to `gatsby-transformer-remark` to tweak the markdown at parse time, rather than after the markdown has already been transformed to HTML at render time.

GuySartorelli:
PR merged

PR #275, opened 2023-10-17
FIX Allow callout and children in code blocks

GuySartorelli:
- Fixes `[CHILDREN]` to be escaped when rendered in a `<code>` block
- Strengthens callout block rendering by doing it the same way children lists are.

## Issues
- https://github.com/silverstripe/doc.silverstripe.org/issues/274
- https://github.com/silverstripe/doc.silverstripe.org/issues/220

### DIFF
--- a/src/utils/parseHTML.ts
+++ b/src/utils/parseHTML.ts
@@ -9,6 +9,7 @@ import { ReactElement } from 'react';
 import rewriteTable from './rewriteTable';
 import rewriteHeader from './rewriteHeader';
 import cleanHeaders from './cleanHeaders';
+import parseCalloutTags from './parseCalloutTags';
 
 /**
  * Replace all the [CHILDREN] with proper React components.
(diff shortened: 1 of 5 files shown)


src/utils/parseHTML.ts
```diff
@@ -38,9 +39,12 @@ const parseHTML = (html: string): ReactElement | ReactElement[] | string => {
                     return rewriteHeader(domNode);
                 }
             }
-            if (domNode.data) {
+            if (domNode.data && domNode.parent?.name !== 'code') {
```
GuySartorelli, line 42: This is the main thing - only render the special elements if they're not inside a code block.

src/utils/parseHTML.ts
```diff
@@ -38,9 +39,12 @@ const parseHTML = (html: string): ReactElement | ReactElement[] | string => {
...
                 const { data } = domNode;
-                return parseChildrenOf(data);
+                if (data.match(/\[CHILDREN.*?\]/)) {
+                    return parseChildrenOf(data);
+                }
+                return parseCalloutTags(data);
             }
```
GuySartorelli, line 48: We now only parse callout tags if they're not inside a code block - and that's why we need to change the way we're rendering them from being just a string replacement to being a proper react component.

emteknetnz (approved):

Issue #276, opened 2023-10-17
Callout blocks aren't rendering sometimes


GuySartorelli:
Looks like my attempts to fix https://github.com/silverstripe/doc.silverstripe.org/issues/220 broke callouts in some scenarios
e.g. on https://docs.silverstripe.org/en/5/developer_guides/model/data_model_and_orm/
![image](https://github.com/silverstripe/doc.silverstripe.org/assets/36352093/26b9c183-aab7-4718-b825-a13dffbb28cc)

## PRs
- https://github.com/silverstripe/doc.silverstripe.org/pull/277

PR #277, opened 2023-10-17
FIX Render callout blocks correctly

GuySartorelli:
- `<p>` tags surrounding callout blocks are swapped out for `<callout>` tags, which are used to identify the element that needs to be replaced.
- If there are paragraphs _inside_ the callout block, the basic regex replace we did will have broken them, so we need to swap them out for `<br>` tags. This renders the same as if they were `<p>` tags because they'll come in pairs.
  - e.g. `<p>[hint]some text</p><p>more text[/hint]</p>` would have become `<callout type="hint">some text</p><p>more text</callout>` which would break the page. So we turn it into `<callout type="hint">some text<br><br>more text</callout>`
- Also protects us from accidentally not closing callout blocks - previously `[hint]some text[hint][info]more text[/info]` would have turned into a single hint callout, because we didn't close the hint block correctly. This has resulted in plenty of errors in the past. Now it will simply not try to render the hint block, so the page won't be _as_ badly broken.

## Issue
- https://github.com/silverstripe/doc.silverstripe.org/issues/276

### DIFF
--- a/src/utils/parseCalloutTags.ts
+++ b/src/utils/parseCalloutTags.ts
@@ -5,19 +5,8 @@ import CalloutBlock from '../components/CalloutBlock';
  * Turn [hint] and other callouts into a proper React component.
  * @param data
  */
-const parseCalloutTags = (data: any): ReactElement|false => {
-    let matches;
-
-    matches = data.match(/\[(hint|warning|info|alert|notice|note)\](.*?)\[\/(?:hint|warning|info|alert|notice|note)\]/s);
-
-    if (matches) {
-        const type = matches[1];
-        const content = matches[2];
-
-        return createElement(CalloutBlock, { type, content });
-    }
-
-    return false;
+const parseCalloutTags = (type: string, content: any): ReactElement|false => {
+    return createElement(CalloutBlock, { type, content });
 };
 
 export default parseCalloutTags;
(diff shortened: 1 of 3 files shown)


src/utils/parseHTML.ts
```diff
@@ -38,13 +38,15 @@ const parseHTML = (html: string): ReactElement | ReactElement[] | string => {
                 if (name.match(/^h[0-9]$/)) {
                     return rewriteHeader(domNode);
                 }
+                if (name === 'callout') {
+                    return parseCalloutTags(domNode.attribs.type, domToReact(domNode.children));
+                }
```
GuySartorelli, line 43: We don't need the `&& domNode.parent?.name !== 'code'` protection anymore, because code blocks won't have rendered the p tag into a callout tag.

src/utils/parseCalloutTags.ts
```diff
@@ -5,19 +5,8 @@ import CalloutBlock from '../components/CalloutBlock';
...
-    if (matches) {
-        const type = matches[1];
-        const content = matches[2];
-
-        return createElement(CalloutBlock, { type, content });
-    }
-
-    return false;
+const parseCalloutTags = (type: string, content: any): ReactElement|false => {
+    return createElement(CalloutBlock, { type, content });
```
GuySartorelli, line 9: We don't need to double up the regex anymore, so this ends up being a lot simpler.

emteknetnz (approved):

PR #286, opened 2024-02-15
ENH Add linkfield docs

GuySartorelli:
Adds linkfield docs to the "optional features" section

Note the preview won't show it, because it's not looking at the creative-commoners PR nor does it have the change from the dev docs PR to add the section link.

## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/27

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -48,7 +48,7 @@ exports.onCreateNode = async ({ node, getNode, getNodesByType, actions, createNo
     }
     return;
   }   
-  const basePath = category === 'user' ? `docs/en/userguide` : `en`;
+  const basePath = category === 'user' ? 'docs/en/userguide' : (thirdparty ? 'docs/en' : 'en');
   const filePath = createFilePath({
     node,
     getNode,
--- a/sources-docs.js
+++ b/sources-docs.js
@@ -11,6 +11,15 @@ module.exports = [
   {
     resolve: 'gatsby-source-git',
     options: {
+      name: 'docs--5--optional_features/linkfield',
+      remote: 'https://github.com/silverstripe/silverstripe-linkfield.git',
+      branch: '4.0',
+      patterns: 'docs/en/**'
+    }
+  },
+  {
+    resolve: 'gatsby-source-git',
+    options: {
       name: 'docs--4',
       remote: 'https://github.com/silverstripe/developer-docs.git',
       branch: '4.13',
--- a/src/utils/rewriteLink.ts
+++ b/src/utils/rewriteLink.ts
@@ -96,7 +96,33 @@ const rewriteLink = (
         )
     }
 
-    // absolute links
+    // absolute links to docs (unversioned)
+    if (href.match(/^https?:\/\/docs.silverstripe.org\/en\/[a-zA-Z]/)) {
+        return createElement(
+            Link,
+            {
+                to: path.join('/', 'en', version, href.replace(/^https?:\/\/docs.silverstripe.org\/en/, '')),
+                className: 'gatsby-link'
+            },
+            domToReact(children, parseOptions)
+        );
+    }
+
+    // absolute links to docs (versioned)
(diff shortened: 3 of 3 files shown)


sources-docs.js
```diff
@@ -8,6 +8,15 @@ module.exports = [
       patterns: 'en/**'
     }
   },
+  {
+    resolve: `gatsby-source-git`,
+    options: {
+      name: `docs--5--optional_features/linkfield`,
```
GuySartorelli, line 14: This name is how we dictate where it'll be put - see `sources-user.js`.

gatsby-node.js
```diff
@@ -48,7 +48,7 @@ exports.onCreateNode = async ({ node, getNode, getNodesByType, actions, createNo
     }
     return;
   }   
-  const basePath = category === 'user' ? `docs/en/userguide` : `en`;
+  const basePath = category === 'user' ? `docs/en/userguide` : thirdparty ? `docs/en` : `en`;
```
GuySartorelli, line 51: `thirdparty` is truthy for any source that has content after `docs--5`.
This base path is used to decide what part of the file path to remove.
Without this change, the linkfield docs end up at `/en/5/optional_features/linkfield/docs/en/` instead of `/en/5/optional_features/linkfield/`

src/utils/rewriteLink.ts
```diff
@@ -96,7 +96,33 @@ const rewriteLink = (
...
+    if (href.match(/^https?:\/\/docs.silverstripe.org\/en\/[a-zA-Z]/)) {
+        return createElement(
+            Link,
+            {
+                to: path.join('/', 'en', version, href.replace(/^https?:\/\/docs.silverstripe.org\/en/, '')),
+                className: 'gatsby-link'
+            },
+            domToReact(children, parseOptions)
+        );
+    }
```
GuySartorelli, line 109: This is needed to rewrite the linkfield docs correctly, with the current version of docs applied. It won't matter immediately, but it _will_ matter when CMS 6 comes around. May as well get it right now instead.

src/utils/rewriteLink.ts
```diff
@@ -96,7 +96,33 @@ const rewriteLink = (
...
+        const hrefVersion = hrefMatch[1];
+        return createElement(
+            Link,
+            {
+                to: path.join('/', 'en', hrefVersion, href.replace(/^https?:\/\/docs.silverstripe.org\/en\/([0-9]+)/, '')),
+                className: 'gatsby-link'
+            },
+            domToReact(children, parseOptions)
+        );
+    }
```
GuySartorelli, line 123: Technically unrelated change but very similar to the above. Sometimes we explicitly want to link to docs from another version (for example in the 5.0.0 changelog we link to CMS 4 upgrade guides).

This rewrites those links to be relative instead of absolute, and doesn't open them in a new tab.

gatsby-node.js
emteknetnz, line 51:
```suggestion
  const basePath = category === 'user' ? 'docs/en/userguide' : (thirdparty ? 'docs/en' : 'en');
```
I'm fairly sure this change makes no difference (it shouldn't :-) though just make this clearer since there's a double ternary

Also change the backticks to single quotes

sources-docs.js
```diff
@@ -8,6 +8,15 @@ module.exports = [
       patterns: 'en/**'
     }
   },
+  {
+    resolve: `gatsby-source-git`,
```
emteknetnz, line 12: Use single quotes not backticks for all of the changes here

emteknetnz (changes requested):

gatsby-node.js
GuySartorelli: Done

sources-docs.js
GuySartorelli: Oops yup good call. Copy pasted from the `sources-user.js` file which uses backticks for some reason.

emteknetnz (approved):

PR #290, opened 2024-04-08
ENH Add elemental docs

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/985

### DIFF
--- a/sources-docs.js
+++ b/sources-docs.js
@@ -23,6 +23,14 @@ module.exports = [
       name: 'docs--5--optional_features/userforms',
       remote: 'https://github.com/silverstripe/silverstripe-userforms.git',
       branch: '6.2',
+    }
+  },
+  {
+    resolve: 'gatsby-source-git',
+    options: {
+      name: 'docs--5--optional_features/elemental',
+      remote: 'https://github.com/silverstripe/silverstripe-elemental.git',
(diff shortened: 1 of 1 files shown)


GuySartorelli (approved):
LGTM, will merge after docs are merged

sources-user.js
```diff
@@ -18,6 +18,15 @@ module.exports = [
...
   },
+  {
+    resolve: `gatsby-source-git`,
+    options: {
+      name: `user--5--optional_features/elemental`,
+      remote: `https://github.com/silverstripe/silverstripe-elemental`,
+      branch: `5.2`,
+      patterns: `docs/en/userguide/**`,
+    }
+  },
```
GuySartorelli, line 29: Just realised this is a duplicate of the `dnadesign/silverstripe-elemental` entry that's already in this file. Please update that entry.

Please also use the `5` branch for user docs. We so rarely update these docs that we can effectively treat the next minor branch as stable.

GuySartorelli (changes requested):

sources-user.js
emteknetnz: Removed, just use the existing one

GuySartorelli (approved):
As before, will merge after docs are merged

PR #291, opened 2024-04-08
ENH Add userform docs

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-userforms/issues/1266

### DIFF
--- a/sources-docs.js
+++ b/sources-docs.js
@@ -20,6 +20,15 @@ module.exports = [
   {
     resolve: 'gatsby-source-git',
     options: {
+      name: 'docs--5--optional_features/userforms',
+      remote: 'https://github.com/silverstripe/silverstripe-userforms.git',
+      branch: '6.2',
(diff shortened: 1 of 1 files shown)


sources-user.js
```diff
@@ -18,6 +18,15 @@ module.exports = [
...
   },
+  {
+    resolve: `gatsby-source-git`,
+    options: {
+      name: `user--5--optional_features/userforms`,
+      remote: `https://github.com/silverstripe/silverstripe-userforms`,
+      branch: `6.2`,
+      patterns: `docs/en/userguide/**`,
+    }
+  },
```
GuySartorelli, line 29:
```suggestion
```
Userforms is already in user help.

GuySartorelli (changes requested):

sources-user.js
emteknetnz: Removed

GuySartorelli (approved):
LGTM, will merge when docs are merged

PR #317, opened 2024-12-04
FIX Update API links URL for new hosting

GuySartorelli:
As per https://github.com/silverstripe/api.silverstripe.org/pull/127 once we've updated hosting to GitHub Pages the search URL will change from `https://api.silverstripe.org/search/lookup` to `https://api.silverstripe.org/search.html`

> [!IMPORTANT]
> **_DO NOT MERGE_** until hosting has swapped over

## Issue
- https://github.com/silverstripe/api.silverstripe.org/issues/122


GuySartorelli:
Not doing this anymore

Issue #278, opened 2023-10-18
Add "copy to clipboard" to code blocks in docs


GuySartorelli:
We're using prismjs for syntax highlighting of code blocks.
There is a plugin for it that adds a copy to clipboard button: https://prismjs.com/plugins/copy-to-clipboard/

## Notes
- We might want to exclude code blocks with no language set, or which have the "text" language set, as these are usually used for things like visually representing folder structures etc which don't need to be copied to anyone's clipboard.
- Looks like we either need to provide a gatsby plugin to hook into the prismjs plugin, or use https://www.npmjs.com/package/gatsby-remark-prismjs-copy-button
    - Either of these will require updating gatsby - so best to leave until we've done https://github.com/silverstripe/doc.silverstripe.org/issues/288

## Acceptance criteria
- [ ] There is a "copy to clipboard" button for all fenced code blocks

## PRs
- https://github.com/silverstripe/doc.silverstripe.org/pull/327

maxime-rainville:
My inclination would be to put them everywhere. I don't really see what the horrible downside would be if someone accidentally copied a folder structure.

GuySartorelli:
No horrible downside, just seems like weird UX to have a button to copy something when there's no value in copying it. But I don't feel strongly.

GuySartorelli:
PR merged

PR #327, opened 2025-05-21
ENH Add copy to clipboard button

emteknetnz:
Issue https://github.com/silverstripe/doc.silverstripe.org/issues/278


GuySartorelli (approved):
Works well in preview.
Note to anyone looking in from the future - Steve did say he tried an existing plugin but it didn't work.

Issue #281, opened 2024-01-07
Size of `<code>` in headings is too small


sabina-talipova:
Some headings are displayed incorrectly. I'm not sure, but if it's correct style, probably we should discuss this with designers.

![Screenshot 2024-01-08 at 10 47 52 AM](https://github.com/silverstripe/developer-docs/assets/87288324/2d5197dd-c79b-4afd-ac80-4359d954028c)

- https://docs.silverstripe.org/en/5/developer_guides/templates/common_variables/. Search for `Children loops` and `Menu loops`. 

## PRs

- https://github.com/silverstripe/doc.silverstripe.org/pull/342

PR #342, opened 2025-09-04
FIX Make code in headings render the correct size

GuySartorelli:
> [!IMPORTANT]
> There are intentionally 2 commits here. DO NOT SQUASH
> The first commit just makes it easier to get set up locally, since the most recent version of gatsby-cli can't be installed with node 12

The CSS for this already existed, but purgecss was ignoring it.

I initially tried getting it to stop ignoring everything for `^h[1-6]$` and their children but that _completely_ changed the way headings were styled - arguably for the better, but that isn't in scope for this issue.

**Before:**
image

**After:**
image

## Issue

- https://github.com/silverstripe/doc.silverstripe.org/issues/281

### DIFF
--- a/src/theme/assets/scss/theme/_base.scss
+++ b/src/theme/assets/scss/theme/_base.scss
@@ -13,11 +13,15 @@ h1, h2, h3, h4, h5, h6 {
 	font-family: 'Poppins', sans-serif;
 	color: $theme-text-color-primary;
 	font-weight: 600;
+}
 
+/*! purgecss start ignore */
+h1, h2, h3, h4, h5, h6 {
 	code[class*="language-"] {
-		font-size: 0.8em;
+		font-size: 0.9em;
 	}
 }
+/*! purgecss end ignore */
 
 code {
 	background: $theme-bg-light;
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ cp .env.development .env.production
 Then, run the build.
 
 ```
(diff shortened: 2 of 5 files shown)


src/theme/assets/scss/theme/_base.scss
```diff
@@ -13,11 +13,15 @@ h1, h2, h3, h4, h5, h6 {
 	font-family: 'Poppins', sans-serif;
 	color: $theme-text-color-primary;
 	font-weight: 600;
+}
 
+/*! purgecss start ignore */
+h1, h2, h3, h4, h5, h6 {
 	code[class*="language-"] {
-		font-size: 0.8em;
+		font-size: 0.9em;
```
GuySartorelli, line 21: 0.8em looked fine when code was included with other text, but looked a bit small when code was on its own in a heading.

src/theme/assets/scss/theme/_base.scss
```diff
@@ -13,11 +13,15 @@ h1, h2, h3, h4, h5, h6 {
 	font-family: 'Poppins', sans-serif;
 	color: $theme-text-color-primary;
 	font-weight: 600;
+}
 
+/*! purgecss start ignore */
```
GuySartorelli, line 18: See https://github.com/FullHuman/purgecss/blob/v2.3.0/docs/whitelisting.md

Tried without the `!` initially but it didn't work.

Issue #347, opened 2025-11-04
Upgrade to node 24


emteknetnz:
Currently running node 12, this needs to be updated to node 24 which is the latest LTS version 

This may require something of a rebuild

New functionality should be similar though does not have to be exactly the same

## Acceptance Criteria
- [ ] Dependencies are updated to node 24
- [ ] Website is functionally similar, within reason
- [x] Guy to raise PRs which remove ToC from changelogs. These should be merged after the main docs site PR has been merged.

## PRs
- https://github.com/silverstripe/doc.silverstripe.org/pull/352
- https://github.com/silverstripe/silverstripe-mfa/pull/626

## Guy PRs

> [!IMPORTANT]
> Don't merge these until the main docs code PR has been merged, since otherwise merging docs will automatically get deployed with the old code.
>
> Note that I suspect it will be easiest to merge all PRs and then merge-up, since e.g. the changelog entries will cause conflicts at each stage. It'll be easier to just do one set of merge-ups at the end rather than resolving those conflicts before merging each PR.

- https://github.com/silverstripe/developer-docs/pull/858 (cms 4)
- https://github.com/silverstripe/developer-docs/pull/859 (cms 5)
- https://github.com/silverstripe/developer-docs/pull/861 (cms 6)
- https://github.com/silverstripe/developer-docs/pull/860 (cms 6.2)
- https://github.com/silverstripe/silverstripe-elemental/pull/1396
- https://github.com/silverstripe/cow/pull/272

PR #352, opened 2025-12-05
Rebuild site with NextJS

emteknetnz:
Issue https://github.com/silverstripe/doc.silverstripe.org/issues/347

Given the size of the pull-request it's probably easiest to look at the files on your local rather than in github


.github/workflows/keepalive.yml
GuySartorelli: Workflows should be kept.

LICENSE
GuySartorelli: ..... we need to keep this lol

README.md
```diff
@@ -1,135 +1,277 @@
...
-## What to update when creating a new pre-release major branch, making a stable major release, or making a major EOL
+# Development with real cloned content
+npm run dev:docs       # Dev server with developer docs on port 9876
+npm run dev:user       # Dev server with user help on port 9876
+npm run clone:docs     # Clone developer docs without starting dev server
+npm run clone:user     # Clone user help docs without starting dev server
 
-When **creating a new major branch for a pre-release major version**
+# Production build & test
+npm run build          # Build & lint (default: docs context, alias for build:docs)
```
GuySartorelli:
```suggestion
npm run build          # Build, lint, and test (default: docs context, alias for build:docs)
```
If it doesn't currently run the tests, it should be updated to run the tests, consistent with every other `build` js script we run.

README.md
```diff
@@ -1,135 +1,277 @@
...
 
-## Installation
+### Environment Variables
 
-To set up a local instance of [doc.silverstripe.org](https://github.com/silverstripe/doc.silverstripe.org):
+Create `.env.local` for local development:
 
-* Clone this repository to an empty directory
+```bash
+# Algolia DocSearch (optional, search disabled if not set)
```
GuySartorelli: Since this is optional, remove the "Optional Environment Variables" heading - all of the environment variables are optional it sounds like.

AGENTS.md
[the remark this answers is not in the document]
GuySartorelli: This is for guiding coding agents for development according to that website (though I did only quickly glance at it).
Whether someone chooses to use a coding agent or not is up to the individual developer, and is the responsibility of said developer to maintain their own workflow.

Since I don't use agentic AI for my development, I don't want to review much less maintain this file.

README.md
```diff
@@ -1,135 +1,180 @@
...
+npm run build:user     # Clone content, build static files and copy images for user help
 
-Once cloned, from the root of the repository, run the command `yarn dev-docs`
-to instantiate a development server. This will consume all of the markdown files in both major release
-branches and allow you to browse the developer documentation site on `http://localhost:8000` by default
-(see the [Gatsby docs](https://www.gatsbyjs.org/docs/) for instructions on customising the port).
+# Testing
+npm test               # Run all tests
+npm run lint           # Run ESLint (must have 0 errors, 0 warnings)
+npm run start          # Serve built static files from out/ directory
```
GuySartorelli: What does starting the dev server (I assume that's what you mean by serving the files) have to do with testing?

README.md
```diff
@@ -1,135 +1,277 @@
...
 
-Once cloned, from the root of the repository, run the command `yarn dev-docs`
-to instantiate a development server. This will consume all of the markdown files in both major release
-branches and allow you to browse the developer documentation site on `http://localhost:8000` by default
-(see the [Gatsby docs](https://www.gatsbyjs.org/docs/) for instructions on customising the port).
+### Build Output
 
-## Building
+- **`.next/`** - Next.js build cache (dev server)
+- **`out/`** - Static HTML export (production builds via `npm run build`), served at root URL
```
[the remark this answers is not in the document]
GuySartorelli: It's not the most standard name for _us_ though, which I'd argue is more important.

GuySartorelli (changes requested):
I haven't looked at any of the code yet and only gave a quick click around. Here's what I've found so far:

1. The link in this block isn't obvious that it's a link:
    image
    Same with the red variants - presumably also the same for the yellow variant but we haven't cut v7 yet so I can't tell.

1. The tab title is missing context (this PR is the top tab, existing docs are the bottom tab)
    image

1. API docs aren't being rewritten correctly:
    image

1. What's this thing and why isn't it mentioned in the readme? I assume it's only in dev mode:
    image

1. When in light mode the toggle looks too much like a settings cog:
    image
    Maybe turn into a slider toggle instead? That way we see both icons at all times so it's clearer that it's light vs dark mode.

1. Mobile menu icon looks weird but that's a nitpick
    imageimage

1. We need a way to control the "On this page" table of contents per page or per section or something - i.e. look at the 6.0.0 changelog table of contents! It's waaaaay too long.

1. There should be no horizontal scrolling. On the 6.0.0 changelog page I have a horizontal scrollbar.

1. The "Included module versions" expandable section seems to not be rendered in changelogs.
    For that matter, are you doing something to remove table of contents from docs? This code shouldn't be choosing what to render or not render, we should remove the ToC from the docs source markdown instead.

1. TOTP auth shouldn't be directly in the "optional features" section - it should be buried under MFA.

README.md
```diff
@@ -1,135 +1,180 @@
...
+npm run clone:docs    # Clone developer docs (or user help with clone:user)
+npm run dev:docs      # Start dev server with cloned docs
 
-- Remove the major from the `PRE_RELEASE` array in the `getVersionMessage` function in `src/utils/nodes.ts`
-- Update the `getDefaultVersion` function's return value to the new stable major in `src/utils/nodes.ts`
-- Update redirects in `netlify.toml`
-- Update `src/pages/index.tsx` to navigate to the new stable major
-- Add the new major to `redirects` in `gatsby-node.js`
+# Or combine in one command:
+npm run dev:docs      # clone + dev (happens automatically)
```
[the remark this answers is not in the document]
emteknetnz: Opps, removed

README.md
```diff
@@ -1,135 +1,177 @@
...
+```bash
+# Serve the built static site in the out/ directory
+npm run start
 ```
 
-### Docker install
+### Full Command Reference
+
+```bash
+npm install            # Install dependencies (use npm ci, not npm update)
```
GuySartorelli:
```suggestion
npm install            # Install dependencies (use npm install, not npm update)
```

GuySartorelli (changes requested):
Still haven't looked at the code itself, this is just from more clicking around:

1. The light/dark toggle is still a bit unclear - this time because it's not obvious which item is currently active.
    current toggle
    Something more like this is what I was thinking - with icons replacing the text. The slider makes it clear which is active currently.
    example slider toggle
1. Links in "On this page" aren't visually obvious that they're links. Should just use the same styles as all other links.
    on this page
1. Using `npm run build:docs && npm run start`, changelog links don't work (they work fine with `npm run dev:docs` - avoiding this sort of problem is one of the reasons I didn't want a separate "dev" mode that builds differently to "prod" mode). For example, clicking on the "5.4.0" changelog link took me to `http://localhost:3000/en/5/changelogs/5.4.0/` which shows this:
    weird page
1. I still have horizontal scrolling on any width between 1024px and 1250px, and anything less than 916px.


## From previous feedback

>> We need a way to control the "On this page" table of contents per page or per section or something - i.e. look at the 6.0.0 changelog table of contents! It's waaaaay too long.
>
> I don't agree, I think the large TOC is actually really good. The changelog's in particular are more like a reference document then anything else, and it seems particularly useful to have that large detailed TOC when doing an upgrade. It also helps just give an idea of what's actually in the very large changelog document.

I don't think the large TOC is good. Everything under "Full list of removed and changed API (by module, alphabetically)" is just going to be ignored, and gives both more visual clutter and more that people have to scroll past. The longer the TOC is, the more likely it is people will just ignore it, which we don't want.

> It also helps just give an idea of what's actually in the very large changelog document.

The top level of the TOC already provides most of the information. Beyond that IMO we need more control.

>> For that matter, are you doing something to remove table of contents from docs? This code shouldn't be choosing what to render or not render, we should remove the ToC from the docs source markdown instead.
>
> Yes there's some code in there to auto remove them. This seems like a a good thing? I don't know why we wouldn't want this. I'd rather not have to do PR's to remove the existing TOC's from all the changelogs

The docs should be rendered as they are, however they are. If there's a list at the top, that list might be TOC or it might not be - making an implementation that's robust enough to be 100% correct about it is hard. I'll happily make the PRs to remove the TOC from existing docs if making those PRs is the only reason you want to keep that code in there.

>> TOTP auth shouldn't be directly in the "optional features" section - it should be buried under MFA.
>
> This one's just too much work to be worth it honestly. Right now all the code expects to put an "optional module" on the "Optional modules" page.
>
> Right now TOTP is the only module that utilises a feature where you can configure where in the tree it shows (i.e. not just on the optional modules page) i.e. https://github.com/silverstripe/doc.silverstripe.org/blob/master/sources-docs.js#L159C9-L159C9
>
> However it looks goofy in CMS 6 docs because it's the only authenticator - https://docs.silverstripe.org/en/6/optional_features/mfa/authenticators/ - this would make more sense if we were planning to support webuthn long term but we're not.

I don't have a way to judge how much work it would be so I can't speak to if it'll be worth it without an estimate of the effort involved.
That said, if you insist on keeping it in the main list, it should be directly below MFA because it's completely useless without MFA - and the `/en/6/optional_features/mfa/authenticators/` page needs to be removed.


## Please also respond to

https://github.com/silverstripe/doc.silverstripe.org/pull/352#discussion_r2617679179

emteknetnz:
> 1. The light/dark toggle is still a bit unclear - this time because it's not obvious which item is currently active.

That slider you've provided looks kind of hard to implement, and has a tiny font which is hard to read. What I've done instead is retain the existing icons, made in the inactive icon grey and increased the icon size

> 2. Links in "On this page" aren't visually obvious that they're links. Should just use the same styles as all other links.

Updated

> 3. Using npm run build:docs && npm run start, changelog links don't work

Issue turned out to be the that `.` in the changelog urls e.g. `/en/6/changelogs/6.0.0` were interpreted as file extensions by `serve`. I've switched it out for `http-server` and things now work as expected

> I don't think the large TOC is good. Everything under "Full list of removed and changed API ...

I've updated the auto TOC to exclude an headings that are inside of an expander section. This significantly reduces the TOC size on changelogs

> The docs should be rendered as they are ... I'll happily make the PRs to remove the TOC from existing docs if making those PRs is the only reason you want to keep that code in there.

The regular changelogs TOC now render underneath the auto TOC.

> That said, if you insist on keeping it in the main list, it should be directly below MFA because it's completely useless without MFA - and the /en/6/optional_features/mfa/authenticators/ page needs to be removed.

The optional feature modules are all listed alphabetically, so there's no way to control where TOTP shows. However I don't think we need to alter the order they show in - MFA already says install TOTP - https://docs.silverstripe.org/en/6/optional_features/mfa/ - and silverstripe/mfa is a composer dep of silverstripe/totp-authenticator.

Seems like all we need to do is remove the "Authenticators" page from the mfa docs

> CI tests are also not running on this PR

Have added, it's now running. I've somewhat based it on the existing ci.yml files e.g. `permissions: {}`, don't run if pull-request is from same github account (I'm not sure why we do this anymore, though there would have been a reason originally)

While it's running in the PR, I'm going to assume that once it's merged then you can't just open a PR where anyone can just specify any code in ci.yml and it's run - presumably it'll just use the ci.yml on the master branch instead.

emteknetnz:
> There is a high severity vulnerability according to npm audit

Looks like that showed up 5 days ago. Have updated

GuySartorelli:
If I'm on a page (e.g. `/en/6/project_governance/repository_management/`) that doesn't exist in v3 docs, and then I select v3 in the version dropdown, it gives a 404. The 404 page has a black background (even though I'm on light mode) with dark grey text.

1. The 404 page should have the same background as every other page
2. I'm on the fence about whether 404 (this PR's behaviour) or kicking you to the homepage (existing behaviour) is correct.
3. The version in the dropdown always shows v6 when the 404 happens.

GuySartorelli:
If I click around the navigation, remembers which navigation sections were previously open and keeps them open.

This means there are more and more open sections the more I click around, so I have to go manually collapse them all again. This should probably only keep the section you're currently in open.

scripts/copy-images.mjs
```diff
@@ -0,0 +1,154 @@
...
+      totalCopied++;
+    }
+  }
+
+  return totalCopied;
+}
+
+function copyStaticAssets(destDir) {
+  const assetsDir = path.join(rootDir, 'assets');
+  const assetFiles = ['favicon.ico', 'logo.svg'];
```
GuySartorelli: Why hardcoded files? Shouldn't we just copy all files from the assets dir?

src/components/SearchBox.tsx
```diff
@@ -1,51 +1,86 @@
...
+    window.location.href = relativeUrl;
+  }, []);
+
+  return (
+    <DocSearch
+      appId={appId}
+      indexName={indexName}
+      apiKey={apiKey}
+      disableUserPersonalization
+      searchParameters={{
```
GuySartorelli: According to my IDE `indexName` and `searchParameters` are deprecated. I won't hold off merging if these are left as they are though.

README.md
GuySartorelli: Please add back the "What to update when creating a new pre-release major branch, making a stable major release, or making a major EOL" section and update it as appropriate.

GuySartorelli (changes requested):
I saw a couple of `SilverStripe` which should be updated to `Silverstripe CMS`.

Other than that, most of the below is from a quick code review. I doubt I'll look at the code any closer than I did for this pass.

emteknetnz:
> Too big of a gap after the last item in a sublist, e.g the gap before "Dependency changes" in the below screenshot from /en/6/changelogs/6.0.0/

That's styling is on the old "manual" TOC, which you've said you were going to create some PR's to remove now that we've got an automatically generated TOC.

> There is a high severity vulnerability according to npm audit

Has been resolved

> The green inside collapsible areas looks a bit weird - maybe just use white instead?

Have changed to very light grey that's almost white.

> There's still a horizontal scroll on /en/6/changelogs/6.0.0/

Fixed

>> Seems like all we need to do is remove the "Authenticators" page from the mfa docs
> Please raise a PR for that.

Done - https://github.com/silverstripe/silverstripe-mfa/pull/626

> The 404 page should have the same background as every other page

Updated

> I'm on the fence about whether 404 (this PR's behaviour) or kicking you to the homepage (existing behaviour) is correct.

404 seems much more correct to me, that's the point of a 404 after all. This means the behaviour between changing the version selector to a missing page, and entering the url directly in the browser act the same

> The version in the dropdown always shows v6 when the 404 happens.

Fixed

> If I click around the navigation, remembers which navigation sections were previously open and keeps them open.

Have removed that behaviour

> I saw a couple of SilverStripe which should be updated to Silverstripe CMS.

Have updated all the instances of it that I found with a text search

scripts/copy-images.mjs
emteknetnz: Have updated to just copy all files in dir

README.md
emteknetnz: Added back in

scripts/clone-docs.mjs
```diff
@@ -0,0 +1,286 @@
...
+        if (optionalFeatureName) {
+          if (context === 'docs' && relPath === 'userguide') {
+            // For docs context, skip userguide
+            continue;
+          }
+          if (context === 'user' && relPath !== 'userguide' && rel === '') {
+            // For user context at root level, only walk into userguide
+            continue;
+          }
+        }
```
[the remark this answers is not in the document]
emteknetnz: Updated

GuySartorelli:
>> Too big of a gap after the last item in a sublist, e.g the gap before "Dependency changes" in the below screenshot from /en/6/changelogs/6.0.0/
>
> That's styling is on the old "manual" TOC, which you've said you were going to create some PR's to remove now that we've got an automatically generated TOC.

That TOC is just an unordered list. The same styling will apply to any unordered list (and presumably ordered lists as well), so it needs to be fixed.

GuySartorelli:
Children blocks have a couple of problems.

<details>
<summary>Click to see screenshots</summary>

Existing docs site:
image

With this PR: 
image

</details>

Using `/en/6/optional_features/mfa/` as an example, there's two problems:
1. A link to `/en/6/optional_features/mfa/userguide/` which should be ommitted (this should be described in the sources json file)
2. The order doesn't match the existing docs site

I note the sort order in the left menu also matches the children list - i.e. the sort order there isn't correct either. I think it's treating folders as always going after files, instead of relying purely on the names of files/folders to select the sort order.

GuySartorelli:
Child boxes also don't expand vertically when they need to, e.g
image
In this example from `/en/6/optional_features/` the "TinyMCE HTML editor" box should be the same height as the Linkfield box.

GuySartorelli (changes requested):
If I'm in a section that's too deeply nested to appear in the side menu, the side menu completely collapses instead of showing keeping the sections open that would lead here.

For example `/en/6/developer_guides/templates/how_tos/navigation_menu/` should keep "Developer Guides > Templates And Views" expanded.

The breadcrumb also seems to exclude the current page - the last item in the breadcrumb on this page is "How To's" which is the parent page.

GuySartorelli:
There's an indent before headings in a `<details>` collapsible block which isn't applied to lists. For example in the 6.0.0 changelog:
image

Either no content should be indented or all should.

GuySartorelli:
In `/en/6/contributing/documentation/#link-to-children` some of the content inside code blocks has accidentally been stripped out.

<details>
<summary>Click to see screenshots</summary>

Current docs site
image

This PR
image

</details>

emteknetnz:
> If I'm in a section that's too deeply nested to appear in the side menu, the side menu completely collapses instead of showing keeping the sections open that would lead here. ... The breadcrumb also seems to exclude the current page 

Both fixed

> There's an indent before headings in a `<details>` collapsible block which isn't applied to lists. For example in the 6.0.0 changelog:

Fixed

> In /en/6/contributing/documentation/#link-to-children some of the content inside code blocks has accidentally been stripped out.

Fixed

GuySartorelli (changes requested):
>> If I'm in a section that's too deeply nested to appear in the side menu, the side menu completely collapses instead of showing keeping the sections open that would lead here. ... The breadcrumb also seems to exclude the current page
> 
> Both fixed

You increased the number of items that can show in the left menu, but you didn't actually fix either of the underlying bugs. I tested by adding a very deeply nested item.

I think 4 (rather than 5 which seems to be what you've set it to now) is an appropriate number of items to allow in the left menu - and if we nest deeper than that, that should work also.

All other problems are fixed from what I can tell so just this last holdout.

Edit: Have split out: https://github.com/silverstripe/doc.silverstripe.org/issues/354

GuySartorelli (approved):
Ready to reploy I guess, lets see what happens :smiling_imp:

GuySartorelli:
Yay everything's green! But the preview doesn't include search still.

GuySartorelli (approved):
Looks good in preview.

Issue #357, opened 2026-01-08
Urls missing version are not redirectng


emteknetnz:
/en/contributing/code should redirect to /en/6/contributing/code
/contributing/code should also redirect to /en/6/contributing/code

## PRs
- https://github.com/silverstripe/doc.silverstripe.org/pull/359

PR #359, opened 2026-01-08
FIX Redirect urls missing en or version prefix

emteknetnz:
Issue https://github.com/silverstripe/doc.silverstripe.org/issues/357

Does client side redirect on the `/en/[version]/` url prefix isn't present

Have double checked there wasn't any existing code that does partial redirection.

The `/app/[version]/not-found.tsx` file was redundant as the `/app/not-found.tsx` acts as a catch-all


src/app/not-found.module.css
```diff
@@ -0,0 +1,39 @@
+.main {
```
emteknetnz: Unreleated refactoring - move the inline styling on not-found.tsx (the 404 page) to a css file

GuySartorelli (approved):
LGTM

Issue #379, opened 2026-02-19
Table of Contents not building correctly


GuySartorelli:
The [6.2.0-beta1](https://deploy-preview-378--ss-docs.netlify.app/en/6/changelogs/beta/6.2.0-beta1/) changelog preview shows a broken table of contents.

<img width="466" height="310" alt="Image" src="https://github.com/user-attachments/assets/a6318a7b-f10b-4e86-82bf-db0b5be1eb32" />

There's two things going on here.

1. ~~The items in accessibility improvements aren't showing (I think that's because of an intentional limit to how many levels deep the ToC goes - if so, we should move this outside the features section)~~ - dealt with in https://github.com/silverstripe/developer-docs/pull/886
2. Any H2/H3 BELOW accessibility improvements isn't displaying in the ToC

## PRs
- https://github.com/silverstripe/doc.silverstripe.org/pull/381

GuySartorelli:
PR(s) merged.

PR #381, opened 2026-02-19
FIX Prevent inline details and summary tags from affecting TOC

emteknetnz:
Issue https://github.com/silverstripe/doc.silverstripe.org/issues/379

Cause was `<details>` / `<summary>` text inside the contents of the changelog

I tested manually by adding the 5.2.0-beta1 changelog to my local content cache and it rendered the TOC correctly

### DIFF
--- a/src/lib/toc/extract-headings.ts
+++ b/src/lib/toc/extract-headings.ts
@@ -26,8 +26,42 @@ const HEADING_REGEX = /^(#{2,3})\s+(.+?)(?:\s*\{#([^}]+)\})?\s*$/gm;
  * Helper to determine if a position in the markdown is inside a <details> block
  * Tracks all <details> opening and closing tags up to the given position
  */
+function stripCodeFromMarkdown(markdown: string): string {
+  const lines = markdown.split('\n');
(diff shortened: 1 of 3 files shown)


src/lib/toc/extract-headings.ts
```diff
@@ -26,8 +26,42 @@ const HEADING_REGEX = /^(#{2,3})\s+(.+?)(?:\s*\{#([^}]+)\})?\s*$/gm;
  * Helper to determine if a position in the markdown is inside a <details> block
  * Tracks all <details> opening and closing tags up to the given position
  */
+function stripCodeFromMarkdown(markdown: string): string {
```
ChloeSartorelli, line 29: I feel like this will break eventually since we're just using regex to effectively parse markdown... I'd feel more comfortable with it if we could do this all as a plugin to whatever library is parsing the markdown for us, but given all the other regex in this file it's probably too late for that.

I'll just have to accept it and hope it doesn't break :sweat_smile:

GuySartorelli (approved):
LGTM

Issue #380, opened 2026-02-19
Tables cause horizontal scrolling of the page


GuySartorelli:
A table in the CMS 6.2.0-beta1 changelog results in horizontal scrolling of the page. This can be seen in https://deploy-preview-378--ss-docs.netlify.app/en/6/changelogs/beta/6.2.0-beta1/

I can't remember if the old docs site would have wrapped the content within cells or made the table itself scroll - I _think_ the former but I'm not 1000% sure.

## PRs
- https://github.com/silverstripe/doc.silverstripe.org/pull/384

GuySartorelli:
PR(s) merged.

PR #384, opened 2026-02-23
FIX Overflow scroll content tables

emteknetnz:
Issue https://github.com/silverstripe/doc.silverstripe.org/issues/380

### DIFF
--- a/src/components/MarkdownContent.module.css
+++ b/src/components/MarkdownContent.module.css
@@ -25,9 +25,3 @@
   margin-bottom: 0.75rem;
 }
 
-/* Table wrapper for horizontal scrolling on overflow */
-.markdownContent > table {
-  display: block;
-  overflow-x: auto;
-}
-
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -440,6 +440,18 @@ table {
   width: 100%;
   margin: 2rem 0;
   border-collapse: collapse;
+  border: 2px solid var(--theme-divider);
+}
+
+.table-wrapper {
+  width: 100%;
+  max-width: 100%;
+  overflow-x: auto;
+  -webkit-overflow-scrolling: touch;
+}
+
+.table-wrapper table {
+  margin: 0;
 }
 
 table th,
--- a/src/lib/markdown/processor.ts
+++ b/src/lib/markdown/processor.ts
@@ -16,6 +16,7 @@ import { remarkLinks } from './remark-links';
 import { cleanApiTags, setCurrentVersion } from './api-links';
 import { rewriteApiLinksInHtml } from './rewrite-api-links-html';
 import { remarkCustomHeadingIds } from './remark-custom-heading-ids';
+import { rehypeWrapTables } from './rehype-wrap-tables';
 
 /**
  * Custom sanitizer schema that allows api: protocol for links and className on elements
@@ -117,6 +118,7 @@ export async function markdownToHtml(
     .use(highlightCodeBlocks)
     .use(rehypeSlug)
     .use(rehypeAutolinkHeadings, autolinkConfig)
+    .use(rehypeWrapTables)
     .use(rehypeSanitize, sanitizerSchema)
     .use(rehypeStringify);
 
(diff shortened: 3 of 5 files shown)


src/components/MarkdownContent.module.css
```diff
@@ -25,9 +25,3 @@
   margin-bottom: 0.75rem;
 }
 
-/* Table wrapper for horizontal scrolling on overflow */
```
emteknetnz, line 28: I guess this didn't work

ChloeSartorelli: This exact styling DOES work - it's just that the selector is wrong. There doesn't seem to be a `markdownContent` CSS class anywhere in the DOM (there is a `MarkdownContent-module__WHPUXW__markdownContent` CSS class but that's as close as it gets).

The only thing that needs to change is the selector should probably be `main table` - or the `markdownContent` class needs to be correctly applied. We don't need all that markdown processing.

GuySartorelli (changes requested):

src/components/MarkdownContent.module.css
emteknetnz: Here's Gemini's reponse when I asked it "Is setting css display: block on a `<table>` a bad idea?"

tl;dr keep the `<div>` wrapper which this PR added otherwise it's bad for accessibility

<details>
<summary><strong>Click to expand the conversation</strong></summary>
The short answer is: **no, it's not inherently bad**, but it entirely changes how the table behaves. It is a very common technique for responsive design, but it comes with some significant structural and accessibility trade-offs that you need to be aware of.

Here is a breakdown of what happens, when to use it, and what to watch out for.

### What Happens When You Use `display: block`?

By default, a `<table>` has a CSS property of `display: table;`. This tells the browser to use a specific layout algorithm that ensures columns align perfectly across all rows, calculating widths based on the content inside the cells. 

When you change a table to `display: block;`:
* **The grid breaks:** The table stops behaving like a unified grid. The browser treats it like a standard block element (like a `<div>`).
* **Columns lose alignment:** If you also set your `<tr>` or `<td>` elements to `display: block;`, the cells will no longer line up vertically with the cells in the rows above or below them.

---

### When It’s Actually a Good Idea

Setting tables (and their inner elements) to `display: block;` is **one of the most popular ways to make tables responsive on mobile devices.**

If a table has too many columns, it will cause horizontal scrolling or break out of the screen on a mobile phone. To fix this, developers use media queries to change the table, rows, and cells to `display: block;` on small screens. 

```css
@media (max-width: 600px) {
  table, thead, tbody, th, td, tr {
    display: block; 
  }
}
```

---

### The Accessibility (A11y) Catch

This is where `display: block` on a table can be considered "bad." 

When you change the CSS `display` property of semantic table elements (`<table>`, `<tr>`, `<td>`), **some screen readers will strip the table of its semantic meaning.** Instead of announcing "Table with 5 columns and 4 rows," the screen reader might just read it as a bunch of disconnected text blocks. This makes it incredibly difficult for visually impaired users to understand the relationship between the data.

**How to fix the a11y issue:**
If you must use `display: block` for responsive tables, you should explicitly add ARIA roles to maintain the table's semantics:

```html
<table role="table">
  <thead role="rowgroup">
    <tr role="row">
      <th role="columnheader">Header</th>
    </tr>
  </thead>
  <tbody role="rowgroup">
    <tr role="row">
      <td role="cell">Data</td>
    </tr>
  </tbody>
</table>
```

---

### A Safer Alternative: The Scrollable Wrapper

If you want to avoid messing with `display: block` and ARIA roles altogether, the modern standard for responsive tables is to leave the table exactly as it is (`display: table`), but wrap it in a `<div>` that handles the overflow. 

```html
<div class="table-wrapper">
  <table>
    </table>
</div>
```

```css
.table-wrapper {
  overflow-x: auto;
  width: 100%;
}
```
This keeps the tabular layout perfectly intact, maintains full screen-reader accessibility without extra ARIA attributes, and allows mobile users to simply swipe horizontally to see the rest of the columns.
</details>

GuySartorelli:
I wonder if it's worth putting a border around tables, so it's visually more obvious when one has become wide enough to be scrollable?

e.g. 
Screenshot from 2026-02-24 12-03-02

compared with
Screenshot from 2026-02-24 12-02-57

I don't feel strongly so am happy to merge without this, but wanted to suggest it as a possibility, since it isn't immediately clear when glancing at the page that the table without the border is scrollable.

emteknetnz:
Have added the border to `table` - matches the border on the generic `th`

```css
table thead th {
  // ..
  border-bottom: 2px solid var(--theme-divider);
}
```

GuySartorelli (approved):
LGTM

PR #392, opened 2026-03-31
Set metadata based on site type

blueo:
small change to allow meta tags to be set per site type (docs/user/search)


emteknetnz (changes requested):
Add unit tests for:

generatePageMetadata()
generateRootMetadata()

emteknetnz (approved):

